PR #1, opened 2024-09-06
WIP API

andyleiserson:


api.bs
```diff
@@ -97,23 +98,47 @@ New additions to the
...
+association between two classes of events: [=impressions=] and [=conversions=].
+
+An <dfn>impression</dfn>, sometimes called a *source event*, is the
+event to which [=conversion=]s are being attributed. Selection of impression
+events is left to the consumer of the API. Examples include:
+
+*   Displaying an advertisement to a user.
+*   Viewing a particular web page.
+
+A <dfn>conversion</dfn>, sometimes called a *trigger event*, is the
```
martinthomson, line 119:
```suggestion
A <dfn>conversion</dfn> is the
```

api.bs
```diff
@@ -97,23 +98,47 @@ New additions to the
...
+event being attributed to [=impression=]s. Selection of conversion events
+is again left to the consumer of the API. Examples include:
+
+*   Signing up for an account.
+*   Making a purchase.
```
martinthomson, line 124:
```suggestion
*   Making a purchase.
*   Visiting a webpage.
```

api.bs
```diff
@@ -97,23 +98,47 @@ New additions to the
...
 
-At conversion time, information for aggregation is created based on the impressions that were previously stored.
-A site can request that the browser select impressions based on a simple query.
+When an [=impression=] occurs, information about the impression is saved by the
+browser in a write-only store. This includes an identifier for the impression
```
martinthomson, line 127:
```suggestion
browser. This includes an identifier for the impression
```

api.bs
```diff
@@ -97,23 +98,47 @@ New additions to the
...
+and some metadata about the impression, such as whether the impression was an
+ad view or an ad click.
```
martinthomson, line 129:
```suggestion
ad view or an ad click.
```

The write-only piece is negotiable.  I can see why a site couldn't modify its store of impressions, but it's also not really necessary if the reading *by other sites* is controlled properly.

The whole business with "what does the e-Privacy Directive say about storing information" is silly.  It matters, of course, but not every regulatory system will grow a PECR-like piece of baggage.

api.bs
```diff
@@ -142,8 +167,147 @@ The aggregation service:
...
+</pre>
+
+Add:
+* attribution system
+* TTL
+* DP parameters
+
+Questions:
+*   Revisit the set of impression types. Can we get rid of it, and put it in the
+    ad ID? Or generalize to "attribution constraint"?
```
martinthomson, line 206: Yes, let's remove that field.  The value here is explainability, but that's pretty marginal.  It's not like we expect regular users to be looking at what sites have stored.

It's also possible that we might gate click impressions on having a click, but I can't see how to use that information constructively.

api.bs
```diff
@@ -142,8 +167,147 @@ The aggregation service:
...
+};
+</xmp>
+
+## SaveImpression API ## {#save-impression-api}
+
+<pre>
+navigator.privateAttribution.saveImpression({
+  type: "view",                  // either "view" or "click"
+  index: 3,                      // the histogram index for counting this impression
+  ad: "sample-campaign-eijb",    // a unique identifier for the ad placement
```
martinthomson, line 194: @benjaminsavage suggested that we make this a number.  Let's open an issue on that.

api.bs
```diff
@@ -142,8 +167,147 @@ The aggregation service:
...
 TODO
 
+## ListAggregationSystems API ## {#list-aggregation-systems-api}
+
+navigator.privateAttribution.listAggregationSystems()
+
+<xmp class=idl>
+dictionary PrivateAttributionAggregationSystem {
+  required DOMString id;
+};
```
martinthomson, line 185: TODO: Work out whether a PrivateAttributionAggregationSystem (wow, right name, hard to type) needs any additional attributes.

api.bs
```diff
@@ -142,8 +167,147 @@ The aggregation service:
...
+dictionary PrivateAttributionImpressionOptions {
+  PrivateAttributionImpressionType type = "view";
+  required unsigned long index;
+  required DOMString ad;
+  required DOMString target;
+};
+
+[SecureContext, Exposed=Window]
+interface PrivateAttribution {
+  [Throws] undefined saveImpression(DOMString aggregationSystemId, PrivateAttributionImpressionOptions options);
```
martinthomson, line 220:
```suggestion
  [Throws] undefined saveImpression(PrivateAttributionImpressionOptions options);
```

Given that we have a bucket of fields, some of which are mandatory, let's move the aggregation system into the bucket.

Except... We don't need to know which system is being used, do we?  I don't think we care: budgeting happens at the point of conversion, so we probably don't care where conversion data is sent as long as the aggregation system is trustworthy.  But at impression time, we really don't care.  Yet.  In the full API we will.  But the bag of attributes method lets us solve that problem neatly.

api.bs
```diff
@@ -142,8 +167,147 @@ The aggregation service:
...
+</xmp>
+
+Implicit saveImpression API inputs:
+* Timestamp (epoch?)
+* Source site
+
+
+### Operation ### {#save-impression-api-operation}
+
+1. Validate inputs
```
martinthomson, line 231:
```suggestion
To <dfn>save an impression</dfn> given `impressionOptions`:

1. Validate inputs.
```

api.bs
```diff
@@ -142,8 +167,147 @@ The aggregation service:
...
+2. If the private attribution API is not enabled, discard the impression data.
+3. Save the impression to the store.
```
martinthomson, line 233:
```suggestion
2. If the private attribution API is enabled, save the impression to the store.
```

martinthomson: Issue: a site can request that we save information to a store that is owned by a site that we have never visited.  That site does not need to exist for this API to work.  Is that OK?

api.bs
```diff
@@ -142,8 +167,147 @@ The aggregation service:
...
+
+
+navigator.privateAttribution.measureConversion({
+  // the number of buckets in the histogram
+  "size": 20,
+
+  // only consider impressions within the last N days
+  lookbackDays: 30,
+  // the type of impression to match against (if omitted, match either)
+  impression: "view",
```
martinthomson, line 250:
```suggestion
```

api.bs
```diff
@@ -142,8 +167,147 @@ The aggregation service:
...
+## MeasureConversion API ## {#measure-conversion-api}
+
+TODO:
+* Add conversion value
+* Change filter data
+
+
+navigator.privateAttribution.measureConversion({
+  // the number of buckets in the histogram
+  "size": 20,
```
martinthomson, line 245:
```suggestion
  size: 20,
  
  // the value to assign to the histogram index of the impression
  value: 3,

```

api.bs
```diff
@@ -142,8 +167,147 @@ The aggregation service:
...
+  source: ["publisher.example"]
+});
+
+
+<xmp class=idl>
+dictionary PrivateAttributionConversionOptions {
+  required unsigned long histogramSize;
+
+  unsigned long lookbackDays = Infinity;
+  PrivateAttributionImpressionType impression;
```
martinthomson, line 263:
```suggestion
  required DOMString aggregationSystemId;
```
Now, I don't think that this is a string that someone will be happy typing often:

```javascript
let whatsit = await privateAttribution.measureConversion({aggregationSystemId: "joe", ...});
```
That's already unwieldy.

api.bs
```diff
@@ -142,8 +167,147 @@ The aggregation service:
...
+  // a list of possible ad identifiers that can be attributed
+  ads: ["sample-campaign-eijb"],
+  // a list of sites where impressions might have been registered
+  source: ["publisher.example"]
+});
+
+
+<xmp class=idl>
+dictionary PrivateAttributionConversionOptions {
+  required unsigned long histogramSize;
```
martinthomson, line 260:
```suggestion
  required unsigned long histogramSize;
  unsigned long value = 1;
```

api.bs
```diff
@@ -142,8 +167,147 @@ The aggregation service:
...
+
+  unsigned long lookbackDays = Infinity;
+  PrivateAttributionImpressionType impression;
+  sequence<DOMString> ads = [];
+  sequence<DOMString> sources = [];
+};
+
+[SecureContext, Exposed=Window]
+partial interface PrivateAttribution {
+  [Throws] ArrayBufferView measureConversion(DOMString aggregationSystemId, PrivateAttributionConversionOptions options);
```
martinthomson, line 270:
```suggestion
  [Throws] Promise<Uint8Array> measureConversion(PrivateAttributionConversionOptions options);
```
A few things:
1. We really need to make this async because it involves I/O.
2. I think that the general trend is toward providing a Uint8Array for stuff like this.
3. The system ID stuff.

api.bs
```diff
@@ -142,8 +167,147 @@ The aggregation service:
...
+
+## User control and visibility ## {#user-control}
+
+* Users should be able to opt out. Opt out should be undetectable.
+* User ability to view the impression store.
+
+# Implementation Considerations # {#implementation-considerations}
+
+* Management and distribution of values for the following:
+    * Histogram size
```
martinthomson, line 307: Two things on this that are somewhat in tension:

1. When it comes down to it, the browser should really care how big histograms are.  Sites get to do basically whatever with CPU and network resources and it's hardly going to be that bad if a site wants a really big histogram.
2. I don't like the idea of really big histograms that basically make a mockery of our aggregation aspirations.

However, I think that the resolution of that tension is in the MPC (DAP specifically).  We might want to ask aggregation service operators to limit histogram size as some function of the number of reports they aggregate. (I still somewhat like the idea of refusing to return a value if an aggregate is smaller than some threshold, even though it is perverse from a strictly DP interpretation perspective.)  If the aggregator supports more, then the API should not be where limits are applied.  Also, different browsers will have different views on this.

martinthomson (commented):
Lots of comments here, most of which lead to more work.  But this is a good starting point.

api.bs
```diff
@@ -142,8 +167,147 @@ The aggregation service:
...
+
+Questions:
+*   Revisit the set of impression types. Can we get rid of it, and put it in the
+    ad ID? Or generalize to "attribution constraint"?
+
+<xmp class=idl>
+enum PrivateAttributionImpressionType { "view", "click" };
+
+dictionary PrivateAttributionImpressionOptions {
+  PrivateAttributionImpressionType type = "view";
```
martinthomson, line 212:
```suggestion
```

api.bs
```diff
@@ -142,8 +167,147 @@ The aggregation service:
...
+* TTL
+* DP parameters
+
+Questions:
+*   Revisit the set of impression types. Can we get rid of it, and put it in the
+    ad ID? Or generalize to "attribution constraint"?
+
+<xmp class=idl>
+enum PrivateAttributionImpressionType { "view", "click" };
+
```
martinthomson, line 210:
```suggestion
```

api.bs
andyleiserson: The `write-only` term was actually in the original and I also had some doubts about it. I think it's an oversimplification. We should discuss the characteristics of the store somewhere (I think I added a section for that), but it takes more than just saying "write only" to capture the important characteristics.

api.bs
andyleiserson: I'm not sure what you mean? The source site is an implicit argument to the API, isn't it by definition the current site?

api.bs
andyleiserson: #6

api.bs
andyleiserson: How about shortening it to `aggregator`?

api.bs
andyleiserson: If we're using DAP, then doesn't the browser need to know the histogram size to construct the report?

api.bs
andyleiserson: I guess we don't strictly need it here in the early binding case, but given that we expect to have it eventually, it seems simpler to include it. It seems better to encourage users to register an impression for each aggregator than to set a precedent of supporting the model where the same impression can be referenced by multiple different aggregators at conversion time. There isn't a lot of complexity in the impression data right now, but what happens if the aggregators want to set conditions (like a limit on impression TTL, or the structure of ad IDs)?

api.bs
andyleiserson: Is it possible that we can omit the `PrivateAttribution` prefix from these definitions? I'm not sure where all they might flow to and whether there is namespacing other than from the prefix.

I ask because I think `PrivateAttributionImpressionOptions` needs to be further refined as something like `PrivateAttributionImpressionOptionsLevel1` (to specify the bag of attributes for the level 1 API), which is really quite a mouthful.

andyleiserson:
I pushed some additional edits here: https://github.com/andyleiserson/api/compare/api...andyleiserson:api:api-2?expand=1 (will open a new PR after this one is merged, or I can add the changes to this one).

martinthomson:
Generally, I'd be OK with iterating on this.  But given its current state, I'm inclined to merge this one and then you are targeting `main` again.  Let me take another quick look, but I think I'll merge this and try to keep up momentum.

api.bs
martinthomson: Yes, we'll need to know what size to make.  We'll ask the site to give us the size they want.

But we don't need to validate or limit that.  If it is the wrong size, that's their problem.

PR #2, opened 2024-09-09
Histogram explanation

martinthomson:


martinthomson:
There are some conflicts here with #1, but I'm happy to resolve those.

PR #5, opened 2024-09-09
define attribution

eriktaubeneck:


martinthomson (changes requested):
My plan is to include an entire section on attribution, such that the definition here (which would be a repetition) is unnecessary.  In another change, I've put the definition section above this one.

eriktaubeneck:
even better. happy to take a stab at that section if it's unwritten

martinthomson:
https://github.com/private-attribution/api/pull/2/files#diff-721a3a36aef527eb7b18f2b40adf9ec0a981167e3fbc8fd4a3f4d8d96f51ad4dR27

PR #11, opened 2024-09-13
Sketch in last-touch attribution

martinthomson:


api.bs
```diff
@@ -450,20 +454,98 @@ how to handle weeks in which the [=privacy budget=] is insufficient,
...
+that does not have sufficient [=privacy budget=].
+If impressions match from a week
+that does not have enough [=privacy budget=],
+impressions are not matched for any preceding weeks.
+That is, once a week has a matching impression
+and insufficient budget,
+the process will set a value of zero for all histogram buckets.
+
+To <dfn ignore>fill a histogram using last-touch attribution</dfn>,
+given |options|:
```
martinthomson, line 476: This method of writing code is super, super annoying.  This is far from being ready, but it has the bones that I think we need.

api.bs
```diff
@@ -613,15 +617,95 @@ the last (most recent) impression that matches the [[#logic-matching|common matc
...
+
+1.  Initialize |impression| to a null value.
+
+1.  Initialize |value| to |options|.|value|.
```
andyleiserson, line 635:
```suggestion
1.  Initialize |value| to |options|.{{PrivateAttributionConversionOptions/value}}.
```

api.bs
```diff
@@ -613,15 +617,95 @@ the last (most recent) impression that matches the [[#logic-matching|common matc
...
+that does not have sufficient [=privacy budget=].
+If impressions match from a week
+that does not have enough [=privacy budget=],
+impressions are not matched for any preceding weeks.
+That is, once a week has a matching impression
+and insufficient budget,
+the process will set a value of zero for all histogram buckets.
+
+To <dfn ignore>fill a histogram using last-touch attribution</dfn>,
+given |options|:
```
andyleiserson, line 631:
```suggestion
given <a dictionary lt=PrivateAttributionConversionOptions>|options|</a>:
```

api.bs
```diff
@@ -613,15 +617,95 @@ the last (most recent) impression that matches the [[#logic-matching|common matc
...
+
+1.  Initialize |impression| to a null value.
+
+1.  Initialize |value| to |options|.|value|.
+
+1.  Let |now| be the current time.<!-- TODO: cite HRTIME spec -->
+
+1.  For each |week| starting from the current week
+    to the oldest week supported by the [=user agent=]:
```
andyleiserson, line 640: We should define somewhere what a week is. Presumably UTC, but need to choose Sunday or Monday.

api.bs
```diff
@@ -613,15 +617,95 @@ the last (most recent) impression that matches the [[#logic-matching|common matc
...
+    1.  Let |impressions| be the result of invoking [=common matching logic=]
+        with |options|, |week|, and |now|.
+
+    1.  If |impressions| is not empty:
+
+        1.  Retain the value of |week|.
+
+        1.  Set |impression| to the value in |impressions|
+            with the most recent |impression|.timestamp.
+            <!-- TODO define a type for stored impressions -->
```
andyleiserson, line 651: We can make the keys in https://private-attribution.github.io/api/#impression-store-contents linkable, at the cost of the `simpledef` markup getting even uglier. I tested `<dfn lt=stored-timestamp>Timestamp</dfn>` before removing it because it was so ugly. Do you have a different kind of definition in mind? Maybe change it to a `<dl>`?

In the case of major clock adjustments, it is possible that the store could have impressions in the future. Is it worth discussing this case?

api.bs
```diff
@@ -613,15 +617,95 @@ the last (most recent) impression that matches the [[#logic-matching|common matc
...
+
+    1.  If |now| - |lookbackDays| is after |impression|.timestamp,
+        continue the loop.
+
+    1.  If |options|.{{PrivateAttributionConversionOptions/ads}}
+        does not contain |impression|.ad,
+        continue the loop.
+
+    1.  If |options|.{{PrivateAttributionConversionOptions/impressionSites}}
+        does not contain |impression|.source,
```
andyleiserson, line 702:
```suggestion
        does not contain |impression|.impressionSite,
```

api.bs
```diff
@@ -613,15 +617,95 @@ the last (most recent) impression that matches the [[#logic-matching|common matc
 The entire [=conversion value=] (up to the maximum imposed by the [[#dp-budget|privacy budget]])
 is allocated to the histogram bucket that was saved with the impression.
 
+Last touch attribution does not select any impression
+that was saved during a week
+that does not have sufficient [=privacy budget=].
```
andyleiserson, line 622: I had been assuming we could report partial conversion value if there is insufficient but non-zero privacy budget. Since the privacy budget is likely to be chosen by the browser (and may vary between browsers and over time), the conversion site doesn't necessarily know the available budget.

andyleiserson (approved):

api.bs
martinthomson: I have a preference for avoiding that sort of thing, at least until we have discussed it.  It might be worth opening an issue to track that though.

api.bs
martinthomson: The links we can work out later.

The local machine won't have any issue with that sort of time skew.  I hope.  (Modulo time adjustments....)

PR #14, opened 2024-09-19
Security and privacy considerations

andyleiserson:


api.bs
```diff
@@ -631,6 +641,36 @@ as long as <a method for=PrivateAttribution>measureConversion()</a>
...
+
+### Clearing ### {#impression-store-clearing}
+
+A mechanism must be provided to clear the impression store.
+For example, the impression store could be cleared
+upon activation of the control that
+[[#opt-out|disables]] the Private Attribution API.
+It is recommended that any mechanism a user agent provides
+to clear stored browsing data (history, cookies, etc.)
+be extended to cover the impression store.
```
martinthomson, line 653: A possible open issue:

While clearing impressions when the browser clears *all* state is reasonable, what would we do for clearing the state for individual sites?  Right now, if we attribute the stored impressions to the conversion site, we'll be in a pretty good state. I think.   If that site clears state, then it loses impressions.  Does that come with risks?

martinthomson: We should also make it obvious that clearing state from the browser DOES NOT clear privacy budget state.  It might consolidate it though (for instance, it might say that all budgets for weeks prior to the clearing is consumed entirely).

martinthomson (approved):

PR #22, opened 2024-10-01
Post TPAC Update

benjaminsavage:
At TPAC, we discussed a few changes to this spec:
1. Using "filterData" instead of enumerating a list of ad identifiers. For now, the proposal was to use the simplest approach of a single integer.
2. Storing information in the impression store about the script which invoked the saveImpression API, which might be a 3rd party script, and might be within an iframe. This could later be used for impression filtering and in the future, as a breakdown key.
3. Providing the maximum conversion value to the device in the saveImpression call as it will be used to deduct appropriate privacy budget loss.

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -5,6 +5,7 @@ Repository: private-attribution/api
 URL: https://private-attribution.github.io/api/
 Editor: Martin Thomson, w3cid 68503, Mozilla https://mozilla.org/, mt@mozilla.com
 Editor: Andy Leiserson, w3cid 147715, Mozilla https://mozilla.org/, aleiserson@mozilla.com
+Editor: Benjamin Savage, w3cid 114877, Meta https://www.meta.com/, btsavage@meta.com
 Abstract: This specifies a browser API for the measurement of advertising performance.  The goal is to produce aggregate statistics about how advertising leads to conversions, without creating a risk to the privacy of individual web users.  This API collates information about people from multiple web origins, which could be a significant risk to their privacy.  To manage this risk, the information that is gathered is aggregated using an aggregation service that is chosen by websites and trusted to perform aggregation within strict limits.  Noise is added to the aggregates produced by this service to provide differential privacy.
 Status Text: This specification is a proposal that is intended to be migrated to the W3C standards track. It is not a standard.
 Text Macro: LICENSE <a href=http://www.w3.org/Consortium/Legal/2015/copyright-software-and-document>W3C Software and Document License</a>
@@ -453,7 +454,7 @@ that the [=user agent=] record an [=impression=] in the [=impression store=].
 <pre>
 navigator.privateAttribution.saveImpression({
   histogramIndex: 3,
-  ad: "sample-campaign-eijb",
+  filterData: 2,
   conversionSite: "advertiser.example",
   lifetimeDays: 7,
 });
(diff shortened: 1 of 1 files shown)


api.bs
```diff
@@ -482,13 +483,11 @@ The arguments to <a method for=PrivateAttribution>saveImpression()</a> are as fo
     [=impression=] with a subsequent [=conversion=], the [=conversion value=]
     will be added to the histogram bucket identified by this index.
   </dd>
-  <dt><dfn>ad</dfn></dt>
+  <dt><dfn>filterData</dfn></dt>
```
martinthomson, line 486: I wonder whether we might find a different name for this.  Issue?

api.bs
```diff
@@ -601,12 +608,24 @@ The arguments to <a method for=PrivateAttribution>measureConversion()</a> are as
     restrictions are satisfied, this value will be encoded into the [=conversion
     report=].
   </dd>
+  <dt><dfn>maxValue</dfn></dt>
+  <dd>
+    The maximum [=conversion value=] across all contributions included in the aggregation.
+    Together with epsilon, this is used to calibrate the distribution of random noise that
+    will be added to the outcome. It is also used to determine the amount of [=privacy budget=] 
+    to expend on this [=conversion report=].
```
martinthomson, line 616: I wonder whether we should set this to `value` if unspecified.  That means that we'd need something else.  I'd need to research how to do that properly within WebIDL though...  Issue?

api.bs
```diff
@@ -601,12 +608,24 @@ The arguments to <a method for=PrivateAttribution>measureConversion()</a> are as
...
-  <dd>A list of [=ad identifiers=]. Only [=impressions=] having one of the listed identifiers may match this [=conversion=].</dd>
+  <dt><dfn>filterData</dfn></dt>
+  <dd>Only [=impressions=] having a [=filterData=] matching this value will be eligible to match this [=conversion=].</dd>
   <dt><dfn>impressionSites</dfn></dt>
-  <dd>A list of impression sites. Only [=impressions=] recorded by one of the impression sites are eligible to match this [=conversion=].</dd>
+  <dd>A list of impression sites. Only [=impressions=] recorded where the top-level site is on this list are eligible to match this [=conversion=].</dd>
+  <dt><dfn>intermediarySites</dfn></dt>
+  <dd>
+    A list of sites which called the [=saveImpression=] API. Only [=impressions=] recorded by scripts originating from one of the 
+    intermediary sites are eligible to match this [=conversion=].
```
martinthomson, line 627: We need to connect this to the "iframe site domain" piece better than this does presently.  Issue?

api.bs
```diff
@@ -639,8 +659,9 @@ The [=impression store=] must store the following information:
 
 <div link-for=PrivateAttribution>
 <pre class=simpledef>
-Ad: The [=/ad identifier=] passed to <a>saveImpression()</a>.
+FilterData: The [=filterData=] passed to <a>saveImpression()</a>.
```
martinthomson, line 662:
```suggestion
Filter Data: The [=filterData=] value passed to <a>saveImpression()</a>.
```

martinthomson (approved):
A few holes, but this is good.

api.bs
benjaminsavage: #26

PR #28, opened 2024-10-02
Update algorithm for budget deduction

bmcase:
Update the algorithm for budget deduction to more closely follow the paper.  Especially as it relates to using the sensitivity to compute the privacy loss.


bmcase:
@martinthomson  updated this PR to also only specify the L1-norm instead of p-norm.

bmcase:
@martinthomson  can we go ahead and merge this PR? I think this is probably good for the budget section for now -- keeping it aligned with the L1 case of the paper.

martinthomson (commented):
I think that we'll need something better integrated into the attribution logic for this to be workable.

bmcase:
We can close this PR as a later PR https://github.com/w3c/ppa/pull/76 makes similar changes to align the spec with the paper.

PR #29, opened 2024-10-02
Flush out aggregation section

bmcase:
add some more detail to MPC aggregation section, DP mechanisms, and encryption of conversion reports.


api.bs
```diff
@@ -883,13 +895,17 @@ TODO
 
 ## Conversion Report Encryption ## {#encryption}
 
-TODO
+In addition to encrypting the histogram contribution, some data is bound to the report as authenticated data.  The authenticated data includes
+1.  The `requested_epsilon` and the `query_global_sensitivity` and the `p-norm` used to enforced the sensitivity.
```
martinthomson, line 899:
```suggestion
1.  The `requested_epsilon` and the `query_global_sensitivity` and the `p-norm` used to compute the sensitivity.
```

Do we really need an arbitrary p-norm, or is it enough to use a straightforward L1 norm?

api.bs
```diff
@@ -1169,11 +1185,17 @@ a safety limit to be exceeded.
 
 ## Differential Privacy Mechanisms ## {#dp-mechanism}
 
-The specific mechanisms that are used
-depend on the type of [=aggregation service=].
+ The two main DP mechanisms support are the Laplace and Gaussian mechanisms.
 
+ For supporting the Laplace mechanism the reports in a query will have all specified the same
```
martinthomson, line 1190:
```suggestion
For supporting the Laplace mechanism the reports in a query will have all specified the same
```

api.bs
```diff
@@ -868,13 +869,24 @@ if the user has opted out of collection of diagnostic data.
 An <dfn>aggregation service</dfn> takes multiple pieces of attribution information
 and produces an aggregate metric.
 
-Each browser will have different requirements for aggregation.
+Each browser will have different requirements for aggregation.  The two aggregation service backends considered here
+are for using a two-party Multi-Party Computation (MPC) or a Trusted Execution Environment (TEE).  In either case the
+Advertiser server (or their delegate) will collect a batch of conversion reports and submit them for aggreation (e.g.
```
martinthomson, line 874:
```suggestion
Advertiser server (or their delegate) will collect a batch of conversion reports and submit them for aggreation (e.g.,
```

martinthomson:
```suggestion
Advertiser server (or their delegate) will collect a batch of conversion reports and submit them for aggreation, such as
```

api.bs
```diff
@@ -868,13 +869,24 @@ if the user has opted out of collection of diagnostic data.
 An <dfn>aggregation service</dfn> takes multiple pieces of attribution information
 and produces an aggregate metric.
 
-Each browser will have different requirements for aggregation.
+Each browser will have different requirements for aggregation.  The two aggregation service backends considered here
+are for using a two-party Multi-Party Computation (MPC) or a Trusted Execution Environment (TEE).  In either case the
```
martinthomson, line 873: Hyperlinks might be a good idea here.  I think that is...

```suggestion
are for using a two-party [[#mpc|Multi-Party Computation (MPC)]] or a [[#tee|Trusted Execution Environment (TEE)]].  In either case the
```

martinthomson (commented):
I have some reservations about the discussion of norms.  It seems like something that doesn't need to be defined so loosely.  We can be more direct than that.

Given the current shape of the API, I don't really see much point in adding anything related to L2 norms, either.

api.bs
bmcase: Yes, we can start with just the L1 norm for simplicity.

bmcase: will update here and in #28

martinthomson:
Hi @bmcase, this seems like it is a bit outdated.  Did you want to keep going, or are you happy for me to close it?

PR #39, opened 2024-10-03
Iterate on aggregator selection

martinthomson:
Use of `setlike` seems to be the right general approach for this, even if the spelling is a bit more verbose than is ideal.

I've filled in more of the details of the protocol and improved some of the linking and references.

With apologies to @bmcase, who will be affected in #29.


api.bs
```diff
@@ -416,36 +408,55 @@ but that is not required,
...
 
+The <dfn enum>PrivateAttributionProtocol</dfn> describes the submission protocol
+used by different [=aggregation services=].  This document defines two protocols:
+
+<dl dfn-for=PrivateAttributionProtocol dfn-type=enum-value>
+  <dt><dfn>dap-11</dfn></dt>
+  <dd>A DAP-based protocol [[DAP]] that uses [=MPC=]; see [[#s-mpc]].</dd>
+  <dt><dfn>tee-00</dfn></dt>
+  <dd>A protocol for submission to a [=TEE=]; see [[#s-tee]].</dd>
+</dl>
```
andyleiserson, line 458: Maybe `dap-11-histogram`, with the possibility that we might add `dap-11-sketch` in the future? It seems like the report format is an additional variable beyond just the aggregation protocol. (Also see https://github.com/private-attribution/api/pull/10#discussion_r1759219801)

api.bs
```diff
@@ -416,36 +408,55 @@ but that is not required,
 and impressions are not scoped to a single aggregation service.
 
 <xmp class=idl>
+enum PrivateAttributionProtocol { "dap-11", "tee-00" };
```
martinthomson, line 411:
```suggestion
enum PrivateAttributionProtocol { "dap-12-histogram", "tee-00" };
```

andyleiserson (approved):

PR #49, opened 2024-10-24
Update README.md link to account for org move

eriktaubeneck:
with the move into the `patcg` org, the URL of the web version needs to be updated.


eriktaubeneck:
the build errors here should be fixed in https://github.com/patcg/ppa-api/pull/50

benjaminsavage (approved):

PR #53, opened 2024-10-31
First cut of DAP/Prio integration

martinthomson:
I'm struggling with this one.  Though this took long enough because bikeshed is annoying for a neophyte, it feels like this requires more detail.  However, for a first cut, I think that this is enough.

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -868,7 +868,7 @@ if the user has opted out of collection of diagnostic data.
 An <dfn>aggregation service</dfn> takes multiple pieces of attribution information
 and produces an aggregate metric.
 
-User agent implementation will have different requirements for aggregation.
+User agent implementations will have different requirements for aggregation.
 However, the aggregation process has some common elements.
 
 Firstly, user agents will need to be configured with,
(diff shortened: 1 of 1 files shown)


api.bs
```diff
@@ -868,12 +868,78 @@ if the user has opted out of collection of diagnostic data.
 An <dfn>aggregation service</dfn> takes multiple pieces of attribution information
 and produces an aggregate metric.
 
-Each browser will have different requirements for aggregation.
+User agent implementation will have different requirements for aggregation.
```
andyleiserson, line 871:
```suggestion
User agent implementations will have different requirements for aggregation.
```

api.bs
```diff
@@ -868,12 +868,78 @@ if the user has opted out of collection of diagnostic data.
...
+the Prio3L1BoundSum instantiation [[PRIO-L1]]
+of the Prio3 Verifiable Distributed Aggregation Function (VDAF) [[VDAF]].
+
+DAP and the Prio3L1BoundSum instantiation define how a report is prepared,
+encrypted, and submitted for aggregation.
+DAP also defines how an aggregate is obtained
+and what configuration is necessary
+for a user agent to obtain about the aggregation service.
+
+Extensions to DAP [[DAP-EXT]] are necessary:
```
andyleiserson, line 920:
```suggestion
Several extensions to DAP [[DAP-EXT]] are necessary for this application:
```

andyleiserson (approved):

PR #54, opened 2024-12-13
Define two features controlled by permissions policy

andyleiserson:


martinthomson (commented):
I think that we can pull in some of the discussion from the presentation material.  Providing a little bit of context about why the choice is made this way will help people understand, rather than just telling them how it is.

We also need to include this in the two [operation](https://patcg.github.io/ppa-api/#save-impression-api-operation) [sections](https://patcg.github.io/ppa-api/#measure-conversion-api-operation).  Though I'm comfortable not doing that immediately: those sections are presently stubs.

api.bs
```diff
@@ -855,6 +855,33 @@ if the user has opted out of collection of diagnostic data.
...
+    identified by the string "<code><dfn export for="PermissionPolicy"
+    enum-value>measure-conversion</dfn></code>".
+
+The [=policy-controlled feature/default allowlist=] for both of these features is
+<code><a dfn for="default allowlist">*</a></code>.
+
+<p class=note>Having separate permissions for
+<a method for=PrivateAttribution>saveImpression()</a> and
+<a method for=PrivateAttribution>measureConversion()</a> might not be necessary,
+but allows pages that do both to limit resources
```
martinthomson, line 875:
```suggestion
<a method for=PrivateAttribution>measureConversion()</a>
allows pages that do both to limit subresources
```

martinthomson (approved):

PR #56, opened 2024-12-13
Initial detail of HTTP API

andyleiserson:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/ppa/pull/56.html" title="Last updated on Apr 10, 2025, 8:37 AM UTC (dd2b274)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ppa/56/3dfabee...dd2b274.html" title="Last updated on Apr 10, 2025, 8:37 AM UTC (dd2b274)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -385,6 +385,10 @@ it may be useful to examine the supported
 <a attribute for=PrivateAttribution>aggregationServices</a> in deciding
 whether to use the Private Attribution API.
 
+It is also possible to register an impression by including the
+[:Save-Impression:]
+header in the HTTP response when serving a resource to the user agent.
+
 To request a conversion report, a site calls
 <a method for=PrivateAttribution>measureConversion()</a>.
 Before calling this API, a site must
@@ -755,8 +759,13 @@ The arguments to <a method for=PrivateAttribution>measureConversion()</a> are as
   <dt><dfn>intermediarySites</dfn></dt>
   <dd>
     A [=set=] of sites which called the <a method for=PrivateAttribution>saveImpression()</a> API.
-    Only [=impressions=] recorded by scripts originating from one of the [=intermediary sites=]
-    are eligible to match this [=conversion=].
+    [=Impressions=] are only eligible to match this [=conversion=] if either:
+    1.  they were registered by a call to <a method for=PrivateAttribution>saveImpression()</a>
+        in a script originating from one of the intermediary sites, or
+    2.  they were registered by a [:Save-Impression:] header associated with a resource
+        from one of the intermediary sites.
+    <!-- TODO: HTTP registration should also apply to impression sites.
+      -- Interaction between impression and intermediary sites needs to be clarified (#55). -->
   </dd>
   <dt><dfn>logic</dfn></dt>
   <dd>
@@ -1108,6 +1117,21 @@ if the user has opted out of collection of diagnostic data.
 ### Visibility ### {#visibility}
 
 * User ability to view the impression store and past report submissions.
+
+
+# HTTP API # {#http-api}
+
+\`<dfn http-header><code>Save-Impression</code></dfn>\` is a
+[=structured header/dictionary|Dictionary Structured Header=]
+set on a response requesting that the user agent invoke the
+<a method for=PrivateAttribution>saveImpression()</a> API.
+
+<pre class=example id=ex-save-impression-header>
+Save-Impression: conversionSite=advertiser.example;histogramIndex=2;\
+  filterData=12;lifetimeDays=7
+</pre>
+
+TODO need to specify in more detail when and how this header is interpreted
 
 
 # Implementation Considerations # {#implementation-considerations}
@@ -1786,6 +1810,12 @@ urlPrefix: https://html.spec.whatwg.org/; spec: html; type: dfn
 urlPrefix: https://infra.spec.whatwg.org/; spec: infra; type: dfn;
     text: user agent
     text: set; url: #sets
+</pre>
+<pre class=anchors>
+spec:structured header; type:dfn; urlPrefix: https://httpwg.org/specs/rfc9651;
+    text: structured header; url: #name-introduction
+    for: structured header
+        text: dictionary; url: #dictionary
 </pre>
 <pre class=biblio>
 {
--- a/api.bs
+++ b/api.bs
@@ -759,13 +759,8 @@ The arguments to <a method for=PrivateAttribution>measureConversion()</a> are as
   <dt><dfn>intermediarySites</dfn></dt>
   <dd>
     A [=set=] of sites which called the <a method for=PrivateAttribution>saveImpression()</a> API.
-    [=Impressions=] are only eligible to match this [=conversion=] if either:
-    1.  they were registered by a call to <a method for=PrivateAttribution>saveImpression()</a>
-        in a script originating from one of the intermediary sites, or
-    2.  they were registered by a [:Save-Impression:] header associated with a resource
-        from one of the intermediary sites.
-    <!-- TODO: HTTP registration should also apply to impression sites.
-      -- Interaction between impression and intermediary sites needs to be clarified (#55). -->
+    Only [=impressions=] recorded by scripts originating from one of the [=intermediary sites=]
+    are eligible to match this [=conversion=].
   </dd>
   <dt><dfn>logic</dfn></dt>
   <dd>
@@ -1139,26 +1134,23 @@ the {{PrivateAttributionImpressionOptions}} dictionary passed to
   <dt><code>conversion-site</code></dt>
   <dd>
     Value of <a dict-member for=PrivateAttributionImpressionOptions>conversionSite</a>,
-    a [=structured header/string=]. This key is required,
-    and must be a valid domain name.
+    a [=structured header/string=]. This key is required.
   </dd>
   <dt><code>histogram-index</code></dt>
   <dd>
     Value of <a dict-member for=PrivateAttributionImpressionOptions>histogramIndex</a>,
-    an [=structured header/integer=]. This key is required.
+    a non-negative [=structured header/integer=]. This key is required.
   </dd>
   <dt><code>filter-data</code></dt>
   <dd>
-    Value of <a dict-member for=PrivateAttributionImpressionOptions>filterData</a>.
-    Implementations parsing the header must accept any valid
-    [=structured header/item=] or [=structured header/inner list=].
-    If the filter data is not understood, no impression is saved.
-    This key is optional.
+    Value of <a dict-member for=PrivateAttributionImpressionOptions>filterData</a>,
+    a non-negative [=structured header/integer=]. This key is optional.
+    If not supplied, the filter data is zero.
   </dd>
   <dt><code>lifetime-days</code></dt>
   <dd>
     Value of <a dict-member for=PrivateAttributionImpressionOptions>lifetimeDays</a>,
-    an [=structured header/integer=]. This key is optional.
+    a positive [=structured header/integer=]. This key is optional.
   </dd>
 </dl>
 
--- a/api.bs
+++ b/api.bs
@@ -1122,7 +1122,7 @@ set on a response requesting that the user agent invoke the
 <a method for=PrivateAttribution>saveImpression()</a> API.
 
 <pre class=example id=ex-save-impression-header>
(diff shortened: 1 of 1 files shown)


api.bs
```diff
@@ -463,7 +467,7 @@ navigator.privateAttribution.saveImpression({
 <xmp class=idl>
 dictionary PrivateAttributionImpressionOptions {
   required unsigned long histogramIndex;
-  required unsigned long filterData;
+  unsigned long filterData;
```
martinthomson, line 470: Do you want to make this a separate change?  Making this optional is probably fine, I'd like that to be an explicit choice.

api.bs
```diff
@@ -855,6 +864,21 @@ if the user has opted out of collection of diagnostic data.
...
 
+# HTTP API # {#http-api}
+
+\`<dfn http-header><code>Save-Impression</code></dfn>\` is a
+[=structured header/dictionary|Dictionary Structured Header=]
+set on a response requesting that the user agent invoke the
+<a method for=PrivateAttribution>saveImpression()</a> API.
+
+<pre class=example id=ex-save-impression-header>
+Save-Impression: conversionSite=advertiser.example;histogramIndex=2;\
```
apasel422, line 875: Is it intentional that this is a single-key dictionary (`conversionSite`) with parameters for `histogramIndex`, `filterData`, and `lifetimeDays`?

Not suggesting it's wrong, but just making sure the intention wasn't for each of those last three to themselves be a top-level key, in which case the `;` separator should be replaced with a space. Making them top-level keys is arguably more flexible, as they can themselves gain parameters over time, but it depends on what future changes, if any, we want this header to support. It's unclear to me if there's a benefit to making this a dictionary (rather than [a single item](https://httpwg.org/specs/rfc9651#item)) if there will only ever be one top-level key.

Related, would it make sense to allow multiple impressions to be registered in a single HTTP response? If so, maybe this should be a [top-level list](https://httpwg.org/specs/rfc9651#list) instead.

Finally, per https://httpwg.org/specs/rfc9651#rfc.section.3.2:

> Member keys cannot contain uppercase characters.

So these might be more conventionally spelled like `conversion-site`. This [also applies to parameters](https://httpwg.org/specs/rfc9651#rfc.section.3.1.2).

api.bs
andyleiserson: I intended this to be a typographical correction because it's already described as optional in the text that follows. But even if so, I agree it's better to consider separately from this PR.

andyleiserson: https://github.com/patcg/ppa-api/pull/57 for making it optional here. I also filed an issue related to the type, https://github.com/patcg/ppa-api/issues/58.

api.bs
andyleiserson: Thanks @apasel422 for reviewing. I've tentatively changed it to commas (i.e. a top-level dictionary) so the definition is at least valid, but depending on the multiple impressions question and the necessary filterData expressiveness, we might want something else. I also added the multiple impressions question as a topic for the meeting.

martinthomson: A top-level dictionary means that we will only permit a single impression to be saved.  It is possible to add *some* structure to parameters like `list=(this is a list of seven items)` but the syntax is a bit unnatural, but it might be that those limitations will encourage good design... or at least creative design.

api.bs
```diff
@@ -855,6 +864,52 @@ if the user has opted out of collection of diagnostic data.
...
+  <dd>
+    Value of <a dict-member for=PrivateAttributionImpressionOptions>histogramIndex</a>,
+    an [=structured header/integer=]. This key is required.
+  </dd>
+  <dt><code>filter-data</code></dt>
+  <dd>
+    Value of <a dict-member for=PrivateAttributionImpressionOptions>filterData</a>.
+    Implementations parsing the header must accept any valid
+    [=structured header/item=] or [=structured header/inner list=].
+    If the filter data is not understood, no impression is saved.
```
apasel422, line 900: Are these sentences in conflict?

And what are the implications of the first sentence for how [filter-data matching](https://patcg.github.io/ppa-api/#logic-matching) works?

andyleiserson: Yeah, it doesn't make a lot of sense as written. I was trying to allow for extensibility of the filter data in the future, but not saving an impression sounds a lot like rejecting the filter data.

martinthomson: I think that we can probably constrain the filter data in both the API and here to avoid the need to allow for problems.  There's a future-proofing argument to be had, but perhaps the way to deal with that is that we add a new dictionary memory (API) or key (HTTP) if we want to change what can be stored with an impression.  Right now, [`filterData`]() is an `unsigned long`, so maybe the answer is to define this as an integer, which is set to zero if the key is absent or the key contains a value that does not fit into the value space of `unsigned long`.

api.bs
```diff
@@ -855,6 +864,52 @@ if the user has opted out of collection of diagnostic data.
...
+</pre>
+
+The following keys are defined, corresponding to the members of
+the {{PrivateAttributionImpressionOptions}} dictionary passed to
+<a method for=PrivateAttribution>saveImpression()</a>.
+
+<dl dfn-for=save-impression>
+  <dt><code>conversion-site</code></dt>
+  <dd>
+    Value of <a dict-member for=PrivateAttributionImpressionOptions>conversionSite</a>,
```
apasel422, line 886: For each of these fields, we should clarify what structured header types they permit.

Based on the examples, I see this as:

- `conversion-site` is a [token](https://httpwg.org/specs/rfc9651.html#token)
- `histogram-index` is a non-negative [integer](https://httpwg.org/specs/rfc9651.html#integer)
- `filter-data` is a non-negative integer (per discussion in https://github.com/w3c/ppa/pull/56/files#r1905845657)
- `lifetime-days` is a positive integer

api.bs
apasel422: +1 to making this an integer for now (we can always accept other types backwards-compatibly in the future), but it's unclear to me if the absence of the field should mean "match any filter data value" or if, as you say, it should be set to `0`, which would make it match only `0`.

CC #57.

api.bs
apasel422: It's possible that we will want `conversion-site` to be a string for generality, given that tokens do have some restrictions on their contents.

api.bs
```diff
@@ -855,6 +864,52 @@ if the user has opted out of collection of diagnostic data.
...
+    a [=structured header/string=]. This key is required,
+    and must be a valid domain name.
```
apasel422, line 888: I'm not sure that `domain name` is precise enough here, based on https://w3c.github.io/ppa/#conversion-site. We might want to say `site` instead.

api.bs
```diff
@@ -855,6 +864,52 @@ if the user has opted out of collection of diagnostic data.
 * User ability to view the impression store and past report submissions.
 
 
+# HTTP API # {#http-api}
+
+\`<dfn http-header><code>Save-Impression</code></dfn>\` is a
+[=structured header/dictionary|Dictionary Structured Header=]
+set on a response requesting that the user agent invoke the
+<a method for=PrivateAttribution>saveImpression()</a> API.
```
apasel422, line 872: Somewhat related to https://github.com/w3c/ppa/pull/106#discussion_r1971560202, do we need to check that the site in Step 2.1 of https://w3c.github.io/ppa/#save-impression-api-operation is using `https://` as its scheme? The JS methods are already marked `[SecureContext]`, but if we need to enforce the use of `HTTPS` then we should probably check that in the core `save impression` and `measure conversion` algorithms for consistency.

martinthomson (approved):

api.bs
```diff
@@ -1110,6 +1114,49 @@ if the user has opted out of collection of diagnostic data.
...
+
+The following keys are defined, corresponding to the members of
+the {{PrivateAttributionImpressionOptions}} dictionary passed to
+<a method for=PrivateAttribution>saveImpression()</a>.
+
+<dl dfn-for=save-impression>
+  <dt><code>conversion-site</code></dt>
+  <dd>
+    Value of <a dict-member for=PrivateAttributionImpressionOptions>conversionSite</a>,
+    a [=structured header/string=]. This key is required.
```
apasel422, line 1137: This is listed as a string, but the example above is encoded as a token. (Note the lack of surrounding `"`.)

@martinthomson Do you think we should bother with string here, or even support both?

apasel422 (changes requested):

api.bs
```diff
@@ -1110,6 +1114,49 @@ if the user has opted out of collection of diagnostic data.
...
+# HTTP API # {#http-api}
+
+\`<dfn http-header><code>Save-Impression</code></dfn>\` is a
+[=structured header/dictionary|Dictionary Structured Header=]
+set on a response requesting that the user agent invoke the
+<a method for=PrivateAttribution>saveImpression()</a> API.
+
+<pre class=example id=ex-save-impression-header>
+Save-Impression: conversion-site=advertiser.example, histogram-index=2, \
+  filter-data=12, lifetime-days=7
```
apasel422, line 1126:
```suggestion
Save-Impression: conversion-site=advertiser.example, histogram-index=2, filter-data=12, lifetime-days=7
```

api.bs
martinthomson: Yeah, I think that we can probably support tokens, but the question is whether we would want to.  We're going to have to deal with IDNA.  Punycode-encoded IDNA will work as a token.  The only reason to stick with just string is to avoid the unnecessary flexibility.

We should be very careful when specifying this, so I would say that we're not done either way.  We need to be very clear that this is punycode-encoded only and that it is not valid to include Unicode, even using the percent-encoded display string type.

api.bs
```diff
@@ -1110,6 +1114,48 @@ if the user has opted out of collection of diagnostic data.
...
 
+# HTTP API # {#http-api}
+
+\`<dfn http-header><code>Save-Impression</code></dfn>\` is a
+[=structured header/dictionary|Dictionary Structured Header=]
+set on a response requesting that the user agent invoke the
+<a method for=PrivateAttribution>saveImpression()</a> API.
+
+<pre class=example id=ex-save-impression-header>
+Save-Impression: conversion-site=advertiser.example, histogram-index=2, filter-data=12, lifetime-days=7
```
martinthomson, line 1125:
```suggestion
Save-Impression: conversion-site="advertiser.example", histogram-index=2, filter-data=12, lifetime-days=7
```

In the spirit of making progress, I think we should follow the advice in RFC 9651 and stick with strings.  Tokens can represent any domain name, so we *could* include tokens or even switch to tokens, but it makes sense to follow the well-lit path on this one.

api.bs
martinthomson:
```suggestion
    a [=structured header/string=].
    The string value includes a domain name using A-labels only;
    [[RFC5890|Internationalized Domain Names]] therefore need to use [[RFC3492|punycode]]. 
    This key is required.
```

Note that this last line is not 100% consistent with the changes in #139.  We'll need to resolve that in #145.

api.bs
```diff
@@ -1110,6 +1114,48 @@ if the user has opted out of collection of diagnostic data.
...
+  <dt><code>histogram-index</code></dt>
+  <dd>
+    Value of <a dict-member for=PrivateAttributionImpressionOptions>histogramIndex</a>,
+    a non-negative [=structured header/integer=]. This key is required.
+  </dd>
+  <dt><code>filter-data</code></dt>
+  <dd>
+    Value of <a dict-member for=PrivateAttributionImpressionOptions>filterData</a>,
+    a non-negative [=structured header/integer=]. This key is optional.
+    If not supplied, the filter data is zero.
```
martinthomson, line 1147:
```suggestion
    If not supplied, a value of 0 is saved for [=impression/Filter Data=].
```

api.bs
```diff
@@ -1110,6 +1114,48 @@ if the user has opted out of collection of diagnostic data.
...
+  </dd>
+  <dt><code>lifetime-days</code></dt>
+  <dd>
+    Value of <a dict-member for=PrivateAttributionImpressionOptions>lifetimeDays</a>,
+    a positive [=structured header/integer=]. This key is optional.
```
martinthomson, line 1152:
```suggestion
    a positive [=structured header/integer=]. This key is optional.
    If not supplied, an [=implementation-defined=] value is saved for [=impression/Lifetime=].
```

martinthomson (approved):
I've made some comments to help move this to a resolution.  In practice, I think that we want an algorithm that calls the ["save an impression"](https://w3c.github.io/ppa/#save-an-impression) API after constructing an `PrivateAttributionImpressionOptions` instance.  However, this has languished long enough already.

PR #57, opened 2024-12-16
Make filterData optional in PrivateAttributionImpressionOptions

andyleiserson:
It is already optional in PrivateAttributionConversionOptions, and described as optional in the detailed description of PrivateAttributionImpressionOptions.

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -474,7 +474,7 @@ navigator.privateAttribution.saveImpression({
 <xmp class=idl>
 dictionary PrivateAttributionImpressionOptions {
   required unsigned long histogramIndex;
-  required unsigned long filterData;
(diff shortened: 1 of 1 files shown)


api.bs
```diff
@@ -474,7 +474,7 @@ navigator.privateAttribution.saveImpression({
 <xmp class=idl>
 dictionary PrivateAttributionImpressionOptions {
   required unsigned long histogramIndex;
-  required unsigned long filterData;
+  unsigned long filterData;
```
martinthomson, line 477: Should we be setting a default or assuming that this ends up as a nullable attribute in any database?  Serious question, because `Option<u64>` and `u64` is significant.

PR #61, opened 2024-12-20
Set a default for `lifetimeDays`

### LINKED ISSUE #45: Set a default for lifetimeDays
There are plenty of cases where a default will work fine.  I suggest 30 days.

andyleiserson:
Resolves #45


martinthomson (approved):
Pick the most sensible default.  There is no one true default here, but this seems fine to me.

PR #71, opened 2025-01-23
Restore a subsection of privacy considerations

andyleiserson:
This text was removed when replacing the ad identifier with filter data. Since I think the consideration remains more or less the same, this change restores the text, modifying it to refer to filter data.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/ppa/pull/71.html" title="Last updated on Feb 20, 2025, 9:52 PM UTC (688a4a3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ppa/71/0609f54...688a4a3.html" title="Last updated on Feb 20, 2025, 9:52 PM UTC (688a4a3)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -1516,24 +1516,25 @@ this difference cannot be detected
 by the site receiving the conversion report.
 
 
-## Placing Identifiers in Filter Data ## {#privacy-filter-data}
+## Including Identifying Information with Saved Impressions ## {#privacy-impression-store}
 
-Because {{PrivateAttributionConversionOptions/filterData}}
-may contain arbitrary information, it is possible to place identifiers
-(e.g. of a user or an advertisement) in this field.
+Sites are able to encode some amount of data
+in impressions,
+using {{PrivateAttributionConversionOptions/filterData}}
+or other fields.
+The API does not prevent sites from encoding user identifiers
+in these fields.
 It is important that identifiers in the filter data cannot be (mis)used
 for arbitrary tracking. The following measures mitigate this risk:
 
 *   The impression store can not be read directly.
-    Thus, identifiers in filter data are only usable for tracking
+    Thus, identifiers are only usable for tracking
     to the extent information about them
     is revealed in [=conversion reports=].
-*   Filter data are scoped
-    to a specific set of impression and conversion sites.
-    It is not possible to use the filter data for tracking across
-    arbitrary sites.
+*   The information in [=conversion reports=] is only revealed
+    after aggregation and the addition of noise.
 *   Users have the ability to [[#impression-store-clearing|clear the impression store]].
-*   Filter data are not written to the impression store
+*   No impressions are saved to the impression store
     when the Private Attribution API is [[#opt-out|disabled]].
 
 
--- a/api.bs
+++ b/api.bs
@@ -1516,6 +1516,27 @@ this difference cannot be detected
 by the site receiving the conversion report.
 
 
+## Placing Identifiers in Filter Data ## {#privacy-filter-data}
+
+Because {{PrivateAttributionConversionOptions/filterData}}
+may contain arbitrary information, it is possible to place identifiers
+(e.g. of a user or an advertisement) in this field.
+It is important that identifiers in the filter data cannot be (mis)used
+for arbitrary tracking. The following measures mitigate this risk:
+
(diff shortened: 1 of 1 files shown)


csharrison:
All of the considerations equally apply to putting identifiers in `histogramIndex` right? Maybe we can combine them here.

api.bs
```diff
@@ -1516,6 +1516,27 @@ this difference cannot be detected
 by the site receiving the conversion report.
 
 
+## Placing Identifiers in Filter Data ## {#privacy-filter-data}
```
martinthomson, line 1519:
```suggestion
## Including Identifying Information with Saved Impressions ## {#privacy-impression-store}
```

api.bs
```diff
@@ -1516,6 +1516,27 @@ this difference cannot be detected
...
+
+Because {{PrivateAttributionConversionOptions/filterData}}
+may contain arbitrary information, it is possible to place identifiers
+(e.g. of a user or an advertisement) in this field.
```
martinthomson, line 1523:
```suggestion
Sites are able to encode some amount of data
in impressions,
using {{PrivateAttributionConversionOptions/filterData}}
or other fields.
The API does not prevent sites from encoding user identifiers
in these fields.
```

api.bs
```diff
@@ -1516,6 +1516,27 @@ this difference cannot be detected
...
+It is important that identifiers in the filter data cannot be (mis)used
+for arbitrary tracking. The following measures mitigate this risk:
```
martinthomson, line 1525:
```suggestion
The data obtained from attribution can use this data
when constructing a [=conversion report=],
which carries some risk of that identifying information
becoming available to the site that receives that report.
The following measures mitigate this risk:
```

api.bs
```diff
@@ -1516,6 +1516,27 @@ this difference cannot be detected
...
+
+*   The impression store can not be read directly.
+    Thus, identifiers in filter data are only usable for tracking
```
martinthomson, line 1528:
```suggestion
    Thus, identifiers are only usable for tracking
```

api.bs
```diff
@@ -1516,6 +1516,27 @@ this difference cannot be detected
...
+    to the extent information about them
+    is revealed in [=conversion reports=].
```
martinthomson, line 1530:
```suggestion
    is revealed in [=conversion reports=].
*   The information in [=conversion reports=] is only revealed
    after aggregation and the addition of noise.
```

api.bs
```diff
@@ -1516,6 +1516,27 @@ this difference cannot be detected
...
+*   Filter data are scoped
+    to a specific set of impression and conversion sites.
+    It is not possible to use the filter data for tracking across
+    arbitrary sites.
```
martinthomson, line 1534:
```suggestion
```

I don't think that this one is particularly relevant.

api.bs
```diff
@@ -1516,6 +1516,27 @@ this difference cannot be detected
...
+*   Users have the ability to [[#impression-store-clearing|clear the impression store]].
+*   Filter data are not written to the impression store
```
martinthomson, line 1536:
```suggestion
*   No impressions are saved to the impression store
```

martinthomson (commented):
I think that this could be made broader than just filterData (whatever we end up renaming it to).

api.bs
andyleiserson: I'm having trouble parsing "The data obtained from attribution can use this data when constructing a [=conversion report=]".

How about:

> The attribution process can use this data when constructing a [=conversion report=], which implies some risk of that identifying information becoming available to the site that receives that report.

api.bs
```diff
@@ -1516,6 +1516,28 @@ this difference cannot be detected
...
+Sites are able to encode some amount of data
+in impressions,
+using {{PrivateAttributionConversionOptions/filterData}}
+or other fields.
+The API does not prevent sites from encoding user identifiers
+in these fields.
+It is important that identifiers in the filter data cannot be (mis)used
+for arbitrary tracking. The following measures mitigate this risk:
+
+*   The impression store can not be read directly.
```
apasel422, line 1530:
```suggestion
*   The impression store cannot be read directly.
```

martinthomson (approved):

Issue #17, opened 2024-09-23
Define our epoch


martinthomson:
> We should define somewhere what a week is. Presumably UTC, but need to choose Sunday or Monday.

_Originally posted by @andyleiserson in https://github.com/private-attribution/api/pull/11#discussion_r1759526857_

martinthomson:
Two obvious options:

1. Take the time since 1970, divide by a week and truncate, each time this increases that's a new epoch.  In other words, the week number is determined by floor(duration since 1970, duration of a week).  We might adjust so that this lines up with a weekend for most people.
2. Start the clock running, for each site, when a site first makes a query.

The first means that there will be cyclical patterns, which might lead to interesting zones for conversions where they are less measured (ideally because they are less interesting, but still).  The second means that there is no guaranteed alignment of cycles and so it might be harder to model bias in results because the bias is distributed across each week.

martinthomson:
@csharrison took an action to look into the privacy consequences of each and whether there are things we need to be careful of.

martinthomson:
Also, there is [this comment](https://github.com/w3c/ppa/issues/78#issuecomment-2649984178) (and the discussion leading to that) which considers adding the ability to target a specific epoch.  That is somewhat harder if your understanding of where an epoch starts and ends is unclear.  That might argue for a fixed epoch cadence.

martinthomson:
@michaelkleber suggests that clients choose a uniformly random week cutoff:

```js
const w = 7 * 24 * 60 * 60 * 1000; 
const week_num = Math.floor(Date.now() / w + Math.random());
```
The random draw would not be a simple random draw, but it would be based on `PRF(secret key, conversion site)` so that the cutover is site-specific and not correlated across sites for the same browser.

martinthomson:
Discussed 2025-03-11, decided to go with https://github.com/w3c/ppa/issues/17#issuecomment-2714559751 for now.  We'll document this to close the PR.

csharrison:
Note that this also is a little more complicated with non-week epoch durations, but should still be workable I think (esp. if the duration divides a week or month, etc)

michaelkleber:
I think the logic is approximately:
* Browser maintains a map from conversion-site to pair<epoch-start-moment, most-recent-budget-spend-moment>.
* If you want to use budget on site X and X is not in the map, then add X to the map with epoch-start-moment = current-time minus uniform_random(0,1) * epoch_length, and most-recent-budget-spend-moment = now.
[...]

PR #72, opened 2025-01-29
Replace hardcoded week with privacy budget epoch

### LINKED ISSUE #70: Refactor the "week" duration to be a global variable
The spec makes extensive use of a hardcoded "week" aspect of the privacy privacy unit, while acknowledging that it is arbitrary. Given that, it seems pretty likely we might get feedback that some other limit is preferrable. This topic came up a bunch in the TPAC WICG breakout ([minutes](https://www.w3.org/2024/09/breakouts/minutes-95.html)).

For now, I think it's probably wise to hardcode this once in a global variable and refer to the variable throughout so any changes can be made seamlessly.

Separately, we might consider if different browsers could set different durations, but that's outside the scope of this issue.

apasel422:
Fixes #70
Relevant to #17

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -750,7 +750,7 @@ enum PrivateAttributionLogic {
 
 Each attribution logic specifies a process for allocating values to histogram buckets.
 This logic includes how to select impressions,
-how to handle weeks in which the [=privacy budget=] is insufficient,
+how to handle [=privacy budget epochs=] in which the [=privacy budget=] is insufficient,
 and (optionally) how to process any additional parameters that might be used.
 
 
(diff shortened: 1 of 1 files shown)


api.bs
```diff
@@ -1168,25 +1168,26 @@ to protect against those attacks.
 ### Privacy Budget Epochs ### {#dp-refresh}
 
 Sites receive a separate differential privacy budget for the data in every
-time internval called and epoch. The epoch length is one week.
+time internval called and epoch. The <dfn>privacy budget epoch</dfn> length is
```
csharrison, line 1171:
```suggestion
time interval called an epoch. The <dfn>privacy budget epoch</dfn> length is
```

csharrison: Feel free to ignore the typo fixed by your other PR

apasel422: CC #73

csharrison:
LGTM

martinthomson (approved):
I like it.  More complicated, but that was inevitable.

Issue #63, opened 2025-01-08
Matching logic is unspecified


apasel422:
Per the TODO at https://patcg.github.io/ppa-api/#logic-matching.

I'd be happy to open a PR to try and address this.

andyleiserson:
PR would be welcome, thanks.

I think the TODO was written before the text that is currently in that section, but the text that's there still isn't sufficient -- the lack of clarity around filter data that you've noted, and it's also missing any discussion of intermediarySites.

PR #74, opened 2025-01-30
Define type for stored impressions and use it in algorithms

apasel422:
CC #63

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/ppa-api/pull/74.html" title="Last updated on Feb 20, 2025, 9:30 PM UTC (bf3ea8f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ppa/74/d632416...apasel422:bf3ea8f.html" title="Last updated on Feb 20, 2025, 9:30 PM UTC (bf3ea8f)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -757,19 +757,18 @@ for=PrivateAttribution>measureConversion()</a> method to find matching
 
 ### Contents ### {#impression-store-contents}
 
-The [=impression store=] must store the following information:
+The [=impression store=] is a [=set=] of [=impression|impressions=]:
 
-<div link-for-hint=PrivateAttribution>
+<div dfn-for=impression link-for-hint=PrivateAttribution>
 <pre class=simpledef>
-Filter Data: The {{PrivateAttributionConversionOptions/filterData}} value passed to <a>saveImpression()</a>.
-Impression Site: The [=impression site=] where <a>saveImpression()</a> was called.
-Intermediary Site: The [=intermediary site=] that called <a>saveImpression()</a>,
+<dfn>Filter Data</dfn>: The {{PrivateAttributionConversionOptions/filterData}} value passed to <a>saveImpression()</a>.
+<dfn>Impression Site</dfn>: The [=impression site=] where <a>saveImpression()</a> was called.
+<dfn>Intermediary Site</dfn>: The [=intermediary site=] that called <a>saveImpression()</a>,
     or `undefined` if the API was invoked by the [=impression site=].
-Conversion Sites: The [=set=] of [=conversion sites=] that were passed to <a>saveImpression()</a>.
-Timestamp: The time at which <a>saveImpression()</a> was called.
-Lifetime: The number of days an [=/impression=] remains eligible for attribution,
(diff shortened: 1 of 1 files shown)


api.bs
```diff
@@ -676,18 +676,17 @@ for=PrivateAttribution>measureConversion()</a> method to find matching
...
 <pre class=simpledef>
-Filter Data: The {{PrivateAttributionConversionOptions/filterData}} value passed to <a>saveImpression()</a>.
-Impression Site: The site that called <a>saveImpression()</a>.
-Intermediary Site: The site corresponding to the script that called <a>saveImpression()</a>.
-Conversion Sites: The conversion site(s) that were passed to <a>saveImpression()</a>.
-Timestamp: The time at which <a>saveImpression()</a> was called.
-Lifetime: The number of days an [=/impression=] remains eligible for attribution,
-Lifetime: either from the call to <a>saveImpression()</a>, or a [=/user agent=]-defined limit.
-Histogram Index: The histogram index passed to <a>saveImpression()</a>.
+<dfn>Filter Data</dfn>: The {{PrivateAttributionConversionOptions/filterData}} value passed to <a>saveImpression()</a>.
```
csharrison, line 683: Do we need to make these all uppercase?

api.bs
```diff
@@ -676,18 +676,17 @@ for=PrivateAttribution>measureConversion()</a> method to find matching
...
+<dfn>Impression Site</dfn>: The site that called <a>saveImpression()</a>.
```
csharrison, line 684:
```suggestion
<dfn>Impression Site</dfn>: The [=site=] that called <a>saveImpression()</a>.
```

csharrison: Ditto below

api.bs
```diff
@@ -676,18 +676,17 @@ for=PrivateAttribution>measureConversion()</a> method to find matching
...
+<dfn>Intermediary Site</dfn>: The site corresponding to the script that called <a>saveImpression()</a>.
+<dfn>Conversion Sites</dfn>: The conversion site(s) that were passed to <a>saveImpression()</a>.
+<dfn>Timestamp</dfn>: The time at which <a>saveImpression()</a> was called.
+<dfn>Lifetime</dfn>: The number of days an [=/impression=] remains eligible for attribution, either from the call to <a>saveImpression()</a>, or a [=/user agent=]-defined limit.
```
csharrison, line 688: Should probably be a `[=duration=]`

api.bs
```diff
@@ -676,18 +676,17 @@ for=PrivateAttribution>measureConversion()</a> method to find matching
...
+<dfn>Histogram Index</dfn>: The histogram index passed to <a>saveImpression()</a>.
```
csharrison, line 689: Can we define a numeric type for this? I think its an `unsigned long long` so this should be a 32 bit unsigned integer.

api.bs
apasel422: We don't need to, but I tried to leave the formatting as is since it was presumably chosen for a reason.

csharrison (approved):
LGTM

PR #80, opened 2025-02-04
Replace erroneous links to "empty" algorithm with "is empty"

apasel422:
["empty"](https://infra.spec.whatwg.org/#list-empty) clears a list, while ["is empty"](https://infra.spec.whatwg.org/#list-is-empty) returns whether the list's size is 0.

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -765,12 +765,12 @@ To <dfn>do attribution and fill a histogram</dfn>, given
 
     1.  If |impressions| is not empty:
 
-        1.  Let |budgetOk| be the result of [=deduct privacy budget=] 
+        1.  Let |budgetOk| be the result of [=deduct privacy budget=]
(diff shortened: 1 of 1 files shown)


csharrison:
Oops, great catch

api.bs
```diff
@@ -786,7 +786,7 @@ To <dfn>do attribution and fill a histogram</dfn>, given
 To <dfn>fill a histogram with last-touch attribution</dfn>, given a [=set=] of
   [=impressions=] |matchedImpressions|, a |histogramSize|, and a |value|:
 
-1. [=assert=] |matchedImpressions| is not [=set/empty=].
+1. [=assert=] |matchedImpressions| is not [=set/is empty|empty=].
```
martinthomson, line 789: Is this better?

```suggestion
1. [=assert=] |matchedImpressions| is [=set/is empty|not empty=].
```

PR #87, opened 2025-02-10
Wrap measureConversion return type in a dictionary for future extensions

### LINKED ISSUE #85: Consider a more future-proof return type for measureConversion
Today [`measureConversion`](https://w3c.github.io/ppa/#dom-privateattribution-measureconversion-options) is defined as follows:

```webidl
[SecureContext, Exposed=Window]
partial interface PrivateAttribution {
  [Throws] Promise<Uint8Array> measureConversion(PrivateAttributionConversionOptions options);
};
```

The `Uint8Array` from the resolved promise is the encrypted report. However, it may be beneficial to future-proof the return type here by introducing a wrapper dictionary. For example, if some unencrypted structured metadata were produced by `measureConversion`, it would be easier to introduce it backwards-compatibly if the definition were instead something like:

```webidl
// exact naming of type and field TBD
dictionary MeasureConversionResult {
  required Uint8Array report;
};

[SecureContext, Exposed=Window]
partial interface PrivateAttribution {
  [Throws] Promise<MeasureConversionResult> measureConversion(PrivateAttributionConversionOptions options);
};
```

This sort of wrapper could also be useful if the API gains additional methods that operate on `MeasureConversionResult`s in their entirety.

apasel422:
Fixes #85

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/ppa-api/pull/87.html" title="Last updated on Feb 10, 2025, 1:51 PM UTC (2ff420f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ppa/87/f05176c...apasel422:2ff420f.html" title="Last updated on Feb 10, 2025, 1:51 PM UTC (2ff420f)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -598,9 +598,13 @@ dictionary PrivateAttributionConversionOptions {
   sequence<DOMString> intermediarySites = [];
 };
 
+dictionary PrivateAttributionConversionResult {
+  required Uint8Array report;
+};
+
 [SecureContext, Exposed=Window]
 partial interface PrivateAttribution {
-  [Throws] Promise<Uint8Array> measureConversion(PrivateAttributionConversionOptions options);
+  [Throws] Promise<PrivateAttributionConversionResult> measureConversion(PrivateAttributionConversionOptions options);
 };
 </xmp>
 


api.bs
```diff
@@ -598,9 +598,13 @@ dictionary PrivateAttributionConversionOptions {
...
+dictionary PrivateAttributionConversionResult {
+  required Uint8Array report;
+};
+
 [SecureContext, Exposed=Window]
 partial interface PrivateAttribution {
-  [Throws] Promise<Uint8Array> measureConversion(PrivateAttributionConversionOptions options);
+  [Throws] Promise<PrivateAttributionConversionResult> measureConversion(PrivateAttributionConversionOptions options);
 };
 </xmp>
```
martinthomson, line 609: This probably needs more.  We don't really describe the return value very well when describing the function, so maybe it is worth doing that down on line 666(old)/670(new) below.

apasel422: I agree, but we're already missing a lot of detail here. For example, `Encrypt the report` and `Return the encrypted report` do not explain how the `Uint8Array` is actually produced, so it might make more sense to fill that in later.

PR #106, opened 2025-02-26
Add procedures for parsing site names from strings 

martinthomson:
This is built on https://github.com/w3c/ppa/pull/104.

This contributes to https://github.com/w3c/ppa/issues/102 by laying some of the ground work regarding
sites.  It should now be possible to "parse a site" when preparing to
invoke the matching logic.  That comes next.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/ppa/pull/106.html" title="Last updated on Apr 24, 2025, 6:18 AM UTC (7f62ed2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ppa/106/eba7f55...7f62ed2.html" title="Last updated on Apr 24, 2025, 6:18 AM UTC (7f62ed2)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -461,9 +461,9 @@ and impressions are not scoped to a single aggregation service.
     "https://aggregator.example/dap",
     "https://example.com/aggregator",
   ];
-  const supportedServices = navigator.privateAttribution?.aggregationServices;
+  const supportedServices = navigator.privateAttribution?.aggregationServices?.values();
   const serviceUrl = preferredServices.find(preference => {
-    supportedServices?.some(svc => svc.url === preference)
+    supportedServices.some(svc => svc.url === preference)
   })?.url;
   </xmp>
 
--- a/api.bs
+++ b/api.bs
@@ -1020,10 +1020,13 @@ and <a dictionary lt=PrivateAttributionConversionOptions>|options|</a>:
              [=obtain a site|obtaining a site=] from the [=origin=].
         <!-- TODO: the intermediary site is currently unused -->
 1. Validate the page-supplied API inputs:
-    1.  If <a dict-member for=PrivateAttributionConversionOptions>logic</a>
-        is specified, and the value is anything other than
-        <a enum-value for=PrivateAttributionLogic>"last-touch"</a>,
-        throw a {{TypeError}}.
+    1.  Switch on the value of <a dict-member for=PrivateAttributionConversionOptions>logic</a>:
+        <dl class="switch">
+        </dl>
+        :   <a enum-value for=PrivateAttributionLogic>"last-touch"</a>
+        ::  If <a dict-member for=PrivateAttributionConversionOptions>value</a>
+            is greater than <a dict-member for=PrivateAttributionConversionOptions>maxValue</a>,
+            throw a {{RangeError}}.
     1. If |options|.{{PrivateAttributionConversionOptions/lookbackDays}} is 0,
         throw a {{RangeError}}.
 1.  Let |report| be an [=create an all-zero histogram|all-zero histogram=].
--- a/api.bs
+++ b/api.bs
@@ -461,10 +461,8 @@ and impressions are not scoped to a single aggregation service.
     "https://aggregator.example/dap",
     "https://example.com/aggregator",
   ];
-  const supportedServices = navigator.privateAttribution?.aggregationServices?.values();
-  const serviceUrl = preferredServices.find(preference => {
-    supportedServices.some(svc => svc.url === preference)
-  })?.url;
+  const supportedServices = navigator.privateAttribution?.aggregationServices;
+  const serviceUrl = preferredServices.find(url => supportedServices?.has(url));
   </xmp>
 
   If the user agent supports the URL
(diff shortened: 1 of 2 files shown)


api.bs
```diff
@@ -833,6 +776,50 @@ to clear stored browsing data (history, cookies, etc.)
...
+The API is therefore able to use a simple [=string=]
+to represent [=sites=].
+
+<p class=note>
+It is also possible for an implementation to internally represent sites
+using just the [=host=] part of the tuple.
+</p>
+
+To <dfn>parse a site</dfn>,
+returning either |site| or failure,
```
apasel422, line 800:
```suggestion
returning either [=site=] or failure,
```
Since otherwise this refers to a local variable called `site`.

api.bs
```diff
@@ -833,6 +776,50 @@ to clear stored browsing data (history, cookies, etc.)
...
+given a [=string=] |input|,
+run these steps:
+
+1.  Assign |host| to the value returned by invoking [=host parser=],
+    passing |input|.
+
+1.  If |host| is failure, return failure.
+
+1.  Assign |site| to the value returned by [=registrable domain|obtain a registrable domain=],
+    passing |host|.
```
apasel422, line 810:
```suggestion
1.  Let |host| be the value returned by invoking [=host parser=],
    passing |input|.

1.  If |host| is failure, return failure.

1.  Let |site| be the value returned by [=registrable domain|obtain a registrable domain=],
    passing |host|.
```
For consistency with our other algorithms.

api.bs
```diff
@@ -833,6 +776,50 @@ to clear stored browsing data (history, cookies, etc.)
...
+### Site Names ### {#site-name-algorithm}
+
+The [=impression store=] saves information
+about three types of [=site=]:
+the [=impression/impression site=],
+an optional [=impression/intermediary site=],
+and a [=set=] of [=impression/conversion sites=].
+
+These [=sites=] MUST all be in [=scheme-and-host=] form,
+with a [=scheme=] of "`https`".
```
apasel422, line 788: I don't think this is a major issue, but do we have any concerns about local development/debugging not being able to use `http://localhost` as a result of making `https` inherent? ARA supports that, for example.

apasel422 (changes requested):

api.bs
martinthomson: Personally, I'm 100% OK with that.  We can talk about what site identity means for localhost, which is a whole other rats nest, but the obvious out there is to use HTTPS and [take steps](https://github.com/FiloSottile/mkcert) to avoid the certificate warnings.

csharrison: IMO we should try to use https://w3c.github.io/webappsec-secure-contexts/#potentially-trustworthy-origin as the check before converting the input to a `site`, which is well-specified and used in many other parts of the platform. That link provides the following justification for local addresses:

> treating such resources as potentially trustworthy is convenient for developers building an application before deploying it to the public.

This should allow `localhost` for user agents that conform to the name resolution rules in [let localhost be localhost](https://datatracker.ietf.org/doc/html/draft-ietf-dnsop-let-localhost-be-localhost). Regarding site identity, we should just be able to use https://html.spec.whatwg.org/multipage/browsers.html#obtain-a-site as-is, which should support localhost.

martinthomson: The problem this is trying to solve is something that "obtain a site" cannot address, because we're looking to turn a string into a site.  That means we _could_ define a parse function for a serialized site, which would also mean that we force people to include "https://" in every string they provide.  Or, as I'm suggesting, we encourage people to setup HTTPS from the outset (`mkcert` really is very convenient) and we can thereby avoid the whole `localhost` mess.

I know that it's not ideal, but local development will require quite a bit more setup than that, because the set of aggregation servers might also need to be overridden (or we might need a developer flag to enable reports that are trivially decoded...).

csharrison: Yeah I was thinking we should just force `https://` and use the URL parser. That seems totally fine with me given that it makes even clearer to callers that the site needs to be secure.

api.bs
```diff
@@ -978,13 +1022,13 @@ and <a dictionary lt=PrivateAttributionConversionOptions>|options|</a>:
...
-                1.  If <a dict-member for=PrivateAttributionConversionOptions>value</a>
-                    is greater than <a dict-member for=PrivateAttributionConversionOptions>maxValue</a>,
-                    throw a {{RangeError}}.
+        :   <a enum-value for=PrivateAttributionLogic>"last-touch"</a>
+        ::  Perform the following steps:
+            1.  If <a dict-member for=PrivateAttributionConversionOptions>value</a> is 0,
+                throw a {{RangeError}}.
+            1.  If <a dict-member for=PrivateAttributionConversionOptions>value</a>
+                is greater than <a dict-member for=PrivateAttributionConversionOptions>maxValue</a>,
+                throw a {{RangeError}}.
```
apasel422, line 1031: Looks like something about the indentation on the following line is causing the build to fail.

apasel422 (changes requested):

api.bs
martinthomson: I hadn't rebased all the changes yet.  I've done so now.

api.bs
```diff
@@ -881,6 +881,50 @@ excludes expired [=impressions=] from [=attribution=]. However, the
...
+is sufficient to identify the site.
+The API is therefore able to use a simple [=string=]
+to represent [=sites=].
+
+<p class=note>
+It is also possible for an implementation to internally represent sites
+using just the [=host=] part of the tuple.
+</p>
+
+To <dfn>parse a site</dfn>,
```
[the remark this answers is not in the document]
martinthomson: Yes, probably.  I've rebased and added a bunch of detail to the saveImpression algorithm.  Hopefully that works.

apasel422 (approved):

PR #112, opened 2025-03-03
Fix erroneous usage of `some` method in example code

apasel422:
The [`PrivateAttributionAggregationServices` interface](https://w3c.github.io/ppa/#privateattributionaggregationservices) is [`setlike`](https://webidl.spec.whatwg.org/#js-setlike), so in order to call `some` we must first call `values`.

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/ppa-api/pull/112.html" title="Last updated on Mar 3, 2025, 5:14 PM UTC (f6ded4d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ppa/112/60a439f...apasel422:f6ded4d.html" title="Last updated on Mar 3, 2025, 5:14 PM UTC (f6ded4d)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -461,9 +461,9 @@ and impressions are not scoped to a single aggregation service.
     "https://aggregator.example/dap",
     "https://example.com/aggregator",
   ];
-  const supportedServices = navigator.privateAttribution?.aggregationServices;
+  const supportedServices = navigator.privateAttribution?.aggregationServices?.values();
   const serviceUrl = preferredServices.find(preference => {
-    supportedServices?.some(svc => svc.url === preference)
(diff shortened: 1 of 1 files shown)


martinthomson:
Would `maplike` be easier to use, or harder?  That is, what difference is there in this example code?

apasel422:
> Would `maplike` be easier to use, or harder? That is, what difference is there in this example code?

To me there's little benefit to this interface being setlike, since the element type is a dictionary, rather than something that can be compared for equality beyond object identity. If it was a maplike, we could have it keyed on the service URL, which would make this example able to do a lookup without calling `some`.

PR #121, opened 2025-03-28
Define epoch

### LINKED ISSUE #17: Define our epoch

martinthomson:
This turned into a bit of an epic, but it implements the changes we
discussed about site-state clearing, epoch start time, and all the
associated logic.

This defines when epochs start (randomly, as agreed).

This defines a store for that per-site state and integrates a new
algorithm for getting the current epoch.

It defines a persistent clear-state time variable that tracks when any
site state was cleared.  This value blocks out use of the API globally
for a period.

I've added notes about partitioning the state-cleared variable, but not
specified anything for that.  That seems like a reasonable balance to
me; as the text says, this is the most private option.

Closes #17.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/ppa/pull/121.html" title="Last updated on Apr 1, 2025, 11:20 PM UTC (e97b746)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ppa/121/8c1f6c2...e97b746.html" title="Last updated on Apr 1, 2025, 11:20 PM UTC (e97b746)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -911,8 +911,7 @@ A <dfn>privacy budget key</dfn> is a [=tuple=] consisting of the following items
 </dl>
 
 To <dfn>deduct privacy budget</dfn>
-given an [=environment settings object=] |settings|,
-a [=privacy budget key=] |key|,
+given a [=privacy budget key=] |key|,
 [=float=] |epsilon|,
 integer |sensitivity|,
 and integer |globalSensitivity|:
@@ -944,8 +943,7 @@ The <dfn>epoch start store</dfn> is a [=map=] keyed by [=site=]
 with values that are [=moments=].
 
 To <dfn>get the current epoch</dfn>
-given an [=environment settings object=] |settings|,
-a [=site=] |site|,
+given a [=site=] |site|,
 and [=moment=] |now|:
 
 1.  Let |period| be the [=duration=] of one [=epoch=].
@@ -1945,8 +1943,6 @@ When clearing site data,
 but retaining browsing history,
 a [=user agent=] can [=map/set=] the value
 in the [=privacy budget store=] to 0 for all [=epochs=]
-up to the [=get the current epoch|current epoch=],
-starting from the [=get the starting epoch for attribution|earliest available epoch=]
 for all affected [=sites=].
 
 
--- a/api.bs
+++ b/api.bs
@@ -948,7 +948,7 @@ given an [=environment settings object=] |settings|,
 a [=site=] |site|,
 and [=moment=] |now|:
 
-1.  Let |period| be the [=duration=] of one [=epoch=] (i.e., 86400 seconds).
+1.  Let |period| be the [=duration=] of one [=epoch=].
 
 1.  If the [=epoch start store=] does not [=map/contain=] |site|, [=map/set=]
     its value to |now|, minus a [=duration=]
@@ -994,7 +994,7 @@ updating the [=last browsing history clear=] can be skipped if:
 
 *   All interactions that are cleared are with [=sites=]
     that have other uncleared interactions.
-*   All cleared interactions occuring during [=epochs=]
+*   All cleared interactions occurring during [=epochs=]
     where other interactions with the same [=site=] were retained.
 *   All affected [=sites=] have a value in the [=epoch start store=].
 
@@ -1003,7 +1003,7 @@ the [=last browsing history clear=] does not need to be updated.
 Instead, the [=privacy budget store=] must be updated
 by [=map/set|setting=] a value of 0
 for all [=privacy budget key=] combinations that can be formed
-from every affected [=sites=] and [=epochs=].
+from every affected [=site=] and [=epoch=].
 
 <p class=note>
 This optimization relies on the [=user agent=]
@@ -1030,12 +1030,12 @@ given [=site=] site:
         This use of [=get the current epoch=]
         can return a negative value.
         This is because the value in [=epoch start store=]
-        is likely to be set to a time be after history was last cleared.
+        is likely to be set to a time after history was last cleared.
 
     1.  Set |clearEpoch| to |clearEpoch| + 2.
 
         <p class=note>
-        Adding <em>two</em> is necessary to that the [=epoch=]
+        Adding <em>two</em> is necessary so that the [=epoch=]
         does not overlap with an [=epoch=]
         before browsing history--
         and the [=epoch start store=]--
@@ -1825,7 +1825,7 @@ In that case, if the API is re-enabled,
 there is no way to determine what budget was expended
 prior to the API being [[#opt-out|disabled]].
 In that case, a [=user agent=] can update
-the value [=last browsing history clear=]
+the value of [=last browsing history clear=]
 to ensure that [=privacy budgets=] are not inadvertently exceeded.
 
 
@@ -1865,7 +1865,7 @@ for a period between one and two [=epochs=]
 from the point that browsing history is discarded.
 Otherwise, a [=site=] that expends its budget
 immediately prior to the clearing of history
-would be able to learn more through API
+would be able to learn more through the API
 than it could without the clearing of state.
 
 <p class=note>
@@ -1879,21 +1879,21 @@ which cannot overlap with any [=epoch=] that might have been defined
 before history was cleared.
 
 A [=user agent=] remembers when history was last cleared
-in the [=last browsing history clear=] variable.
+in the [=last browsing history clear=] value.
 This is used to prevent [=privacy budget|budget=] expenditure
 until it is guaranteed that a [=site=] cannot exceed
 the [=user agent=]-chosen maximum allowed [=privacy budget=].
 
 The [=get the starting epoch for attribution=] algorithm
 ensures that an [=impression site=]
-cannot make queries [=impressions=] from any [=epoch=]
+cannot query [=impressions=] from any [=epoch=]
 that starts within one [=epoch=] [=duration=]
 from when state is cleared.
 
 
 ### Clearing Impressions ### {#removing-impressions}
 
-A mechanism must be provided to clear the impression store.
+A mechanism must be provided to clear the [=impression store=].
 For example, the impression store could be cleared
 upon activation of the control that
 [[#opt-out|disables]] the Private Attribution API.
--- a/api.bs
+++ b/api.bs
@@ -958,7 +958,7 @@ and [=moment=] |now|:
 1.  Let |start| be the result of [=map/get|getting the value=] of |site|
     from the [=epoch start store=].
 
-1.  Let |elapsed| be |now| - |start| / |period|.
+1.  Let |elapsed| be (|now| - |start|) / |period|.
 
 1.  Return |elapsed| as an integer, rounded towards negative Infinity.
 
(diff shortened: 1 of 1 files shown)


api.bs
```diff
@@ -921,6 +932,120 @@ To <dfn>deduct privacy budget</dfn> given a [=privacy budget key=] |key|,
...
+
+The <dfn>epoch start store</dfn> is a [=map=] keyed by [=site=]
+with values that are [=moments=].
+
+To <dfn>get the current epoch</dfn>
+given an [=environment settings object=] |settings|,
+a [=site=] |site|,
+and [=moment=] |now|:
+
+1.  Let |period| be the [=duration=] of one [=epoch=] (i.e., 86400 seconds).
```
apasel422, line 951:
```suggestion
1.  Let |period| be the [=duration=] of one [=epoch=].
```
It might be better to avoid duplicating the actual `86400 second` value here, in case it's changed at some point.

api.bs
```diff
@@ -921,6 +932,120 @@ To <dfn>deduct privacy budget</dfn> given a [=privacy budget key=] |key|,
...
+
+1.  If the [=epoch start store=] does not [=map/contain=] |site|, [=map/set=]
+    its value to |now|, minus a [=duration=]
+    that is randomly selected
+    from between 0 (inclusive) and |period| (exclusive).
+
+1.  Let |start| be the result of [=map/get|getting the value=] of |site|
+    from the [=epoch start store=].
+
+1.  Let |elapsed| be |now| - |start| / |period|.
```
apasel422, line 961: Should this be `(|now| - |start|) / |period|`, i.e. subtraction before division?

api.bs
```diff
@@ -921,6 +932,120 @@ To <dfn>deduct privacy budget</dfn> given a [=privacy budget key=] |key|,
...
+This guarantees that stored data does not reveal anything
+about browsing history.
+</div>
+
+As an optional optimization,
+updating the [=last browsing history clear=] can be skipped if:
+
+*   All interactions that are cleared are with [=sites=]
+    that have other uncleared interactions.
+*   All cleared interactions occuring during [=epochs=]
```
apasel422, line 997:
```suggestion
*   All cleared interactions occurring during [=epochs=]
```

api.bs
```diff
@@ -921,6 +932,120 @@ To <dfn>deduct privacy budget</dfn> given a [=privacy budget key=] |key|,
...
+    where other interactions with the same [=site=] were retained.
+*   All affected [=sites=] have a value in the [=epoch start store=].
+
+If all three conditions are true,
+the [=last browsing history clear=] does not need to be updated.
+Instead, the [=privacy budget store=] must be updated
+by [=map/set|setting=] a value of 0
+for all [=privacy budget key=] combinations that can be formed
+from every affected [=sites=] and [=epochs=].
```
apasel422, line 1006:
```suggestion
from every affected [=site=] and [=epoch=].
```

api.bs
```diff
@@ -921,6 +932,120 @@ To <dfn>deduct privacy budget</dfn> given a [=privacy budget key=] |key|,
...
+    perform the following steps:
+
+    1.  Let |clearEpoch| be the result of [=get the current epoch=]
+        with |site| and the value of [=last browsing history clear=].
+
+        <p class=note>
+        This use of [=get the current epoch=]
+        can return a negative value.
+        This is because the value in [=epoch start store=]
+        is likely to be set to a time be after history was last cleared.
```
apasel422, line 1033:
```suggestion
        is likely to be set to a time after history was last cleared.
```

api.bs
```diff
@@ -921,6 +932,120 @@ To <dfn>deduct privacy budget</dfn> given a [=privacy budget key=] |key|,
...
+
+    1.  Set |clearEpoch| to |clearEpoch| + 2.
+
+        <p class=note>
+        Adding <em>two</em> is necessary to that the [=epoch=]
```
apasel422, line 1038:
```suggestion
        Adding <em>two</em> is necessary so that the [=epoch=]
```

api.bs
```diff
@@ -921,6 +932,120 @@ To <dfn>deduct privacy budget</dfn> given a [=privacy budget key=] |key|,
...
+that arises from a [=privacy budget=] reaching zero
+is averaged across all the contributions from all [=users=].
+
+The <dfn>epoch start store</dfn> is a [=map=] keyed by [=site=]
+with values that are [=moments=].
+
+To <dfn>get the current epoch</dfn>
+given an [=environment settings object=] |settings|,
+a [=site=] |site|,
+and [=moment=] |now|:
```
apasel422, line 949: Do we need to take any special care for the non-monotonically-increasing times throughout these algorithms?

https://w3c.github.io/hr-time/#sec-clocks

api.bs
```diff
@@ -1671,6 +1800,124 @@ of candidate impressions.
...
+from attribution.
+
+A [=user agent=] may clear data
+from the [=privacy budget store=] and [=epoch start store=]
+when the API is [[#opt-out|disabled]].
+In that case, if the API is re-enabled,
+there is no way to determine what budget was expended
+prior to the API being [[#opt-out|disabled]].
+In that case, a [=user agent=] can update
+the value [=last browsing history clear=]
```
apasel422, line 1828:
```suggestion
the value of [=last browsing history clear=]
```

api.bs
```diff
@@ -1671,6 +1800,124 @@ of candidate impressions.
...
+any subsequent [=privacy budget=] expenditure
+cannot cause privacy loss in excess of configured limits
+as a result of losing that information.
+
+This can be achieved by disabling attribution
+for a period between one and two [=epochs=]
+from the point that browsing history is discarded.
+Otherwise, a [=site=] that expends its budget
+immediately prior to the clearing of history
+would be able to learn more through API
```
apasel422, line 1868:
```suggestion
would be able to learn more through the API
```

api.bs
```diff
@@ -1671,6 +1800,124 @@ of candidate impressions.
...
+for affected [=sites=].
+Because the [=epoch start store=] is cleared,
+the start of any newly-defined [=epoch=] for that [=site=]
+becomes unknown.
+Any new [=epoch=] will [start at a fresh randomly-selected time](#s-epoch-start),
+which cannot overlap with any [=epoch=] that might have been defined
+before history was cleared.
+
+A [=user agent=] remembers when history was last cleared
+in the [=last browsing history clear=] variable.
```
apasel422, line 1882:
```suggestion
in the [=last browsing history clear=] value.
```

api.bs
```diff
@@ -1671,6 +1800,124 @@ of candidate impressions.
...
+This is used to prevent [=privacy budget|budget=] expenditure
+until it is guaranteed that a [=site=] cannot exceed
+the [=user agent=]-chosen maximum allowed [=privacy budget=].
+
+The [=get the starting epoch for attribution=] algorithm
+ensures that an [=impression site=]
+cannot make queries [=impressions=] from any [=epoch=]
```
apasel422, line 1889:
```suggestion
cannot query [=impressions=] from any [=epoch=]
```

api.bs
```diff
@@ -1671,6 +1800,124 @@ of candidate impressions.
...
+that starts within one [=epoch=] [=duration=]
+from when state is cleared.
+
+
+### Clearing Impressions ### {#removing-impressions}
+
+A mechanism must be provided to clear the impression store.
```
apasel422, line 1896:
```suggestion
A mechanism must be provided to clear the [=impression store=].
```

apasel422 (changes requested):
Any idea why the "Preview" and "Diff" links are missing from the PR description?

michaelkleber:
Bravo, Martin, thank you for turning that whole discussion in #17 into a spec that all holds together well.

api.bs
martinthomson: Opened #123 for that question, which is separable.

api.bs
martinthomson: Of course it should.

martinthomson (commented):
Thanks for the review.  I have addressed comments and made a tiny change.  I moved the text added to privacy considerations to that bottom of that section, as it is the least consequential of the things we need to discuss there.

apasel422 (changes requested):

apasel422 (approved):

api.bs
```diff
@@ -1780,6 +1908,122 @@ this support is deemed necessary
...
+before history was cleared.
+
+A [=user agent=] remembers when history was last cleared
+in the [=last browsing history clear=] value.
+This is used to prevent [=privacy budget|budget=] expenditure
+until it is guaranteed that a [=site=] cannot exceed
+the [=user agent=]-chosen maximum allowed [=privacy budget=].
+
+The [=get the starting epoch for attribution=] algorithm
+ensures that an [=impression site=]
```
apasel422, line 1994:
```suggestion
ensures that a [=conversion site=]
```
IIUC, it is the conversion site doing the querying, not the impression site.

csharrison:
Sorry I wasn't able to finish my review before the merge (no worries). We can address in a follow-up. Happy to send a PR too.

api.bs
```diff
@@ -921,6 +931,119 @@ To <dfn>deduct privacy budget</dfn> given a [=privacy budget key=] |key|,
...
+at the request of a [=user=].
+The value is initially unset.
+
+<div class=note>
+A [=user agent=] that offers selective deletion of history,
+such as the removal of state for a single site,
+could partition this value so that other sites remain unaffected.
+However, any choice to partition leaks some information
+that might reveal that a specific site had been visited
+to any entity that can access stored data.
```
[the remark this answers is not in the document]
martinthomson: Indeed.  Do you think I should cut this text?

api.bs
```diff
@@ -921,6 +931,119 @@ To <dfn>deduct privacy budget</dfn> given a [=privacy budget key=] |key|,
...
+in the affected [=epochs=].
+This only works because any of the retained interactions
+could have resulted in exhausting the budget.
+
+To <dfn>get the starting epoch for attribution</dfn>
+given [=site=] site:
+
+1.  Let |startEpoch| be a [=user agent=]-defined value
+    for the earliest [=epoch=]
+    that is supported for attribution.
```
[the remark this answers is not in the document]
martinthomson: It depends.  In practice, I would represent an epoch as the difference between now and some fixed time (1970 probably).  Then I would structure these stores as a vector, with the zero offset be the result of [=get the current epoch=], passing in `now - <some span of time>`, storing that first epoch value as an offset.  That is, something like:

```js
{
  store: [1, 2, 3, 4, 5],
  offset: 12,
  index: function(epoch) {
    return this.store[epoch - this.offset];
  }
}
```

The algorithms in the spec will return a negative number in some cases, because the browsing history clear will likely be before the time that the "first" epoch starts for a given site.  That's easily rectified if you stick with a fixed zero point, like 1970.  I just found it easier to explain things without having to define another variable, like this offset.

martinthomson:
No worries.  It's my fault, so I'll take a swing at this and we can discuss at our upcoming meeting.

api.bs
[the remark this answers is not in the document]
csharrison: I'm in favor, I think it kind of muddles things.

Issue #125, opened 2025-04-01
Clear-Site-Data integration


apasel422:
#121 greatly expands on how user-initiated data clearing works, but we should also consider how to integrate with the site-initiated `Clear-Site-Data` header from https://w3c.github.io/webappsec-clear-site-data/, particularly to ensure that sites cannot abuse it in order to exceed the intended budget, while still presenting some utility such as removal of impressions from the impression store prior to their natural expiry, e.g. in case of erroneous metadata.

martinthomson:
This seems straightforward: clearing site data is already addressed, so we can simply refer to that spec.

michaelkleber:
If there was doubt, I agree with Martin's implicit opinion that we should be able to use the easier "Clearing Site Data" hook rather than the more invasive "Clearing Browsing History".

apasel422:
1. Should this deletion correspond to `Clear-Site-Data: storage` or `Clear-Site-Data: cookies`, either, or [something else](https://w3c.github.io/webappsec-clear-site-data/#http-headerdef-clear-site-data)?
2. Given an HTTP response from origin `https://bar.foo.example` that sets the appropriate header, which impressions should be considered for deletion? We currently only retain an `impression site`, not origin, so the obvious behavior would be to match using `same-site`.
And [`Clearing Impressions`](https://w3c.github.io/ppa/#removing-impressions) says:
    > A user agent that clears state for a site must discard all saved impressions that were saved during the affected period with a matching [impression site](https://w3c.github.io/ppa/#impression-impression-site) or [conversion site](https://w3c.github.io/ppa/#impression-conversion-sites).
[...]
Impressions that have a matching [intermediary site](https://w3c.github.io/ppa/#impression-intermediary-site) may be cleared.

    So maybe we also need to look at the impression's `conversion sites` and `intermediary site`?

csharrison:
> This seems straightforward: clearing site data is already addressed, so we can simply refer to that spec.

I don't think that's true, since we probably don't want to give sites the ability to clear their budgets? Right now the `clear-site-data` section mentions setting budgets to 0.

martinthomson:
Oh, you would prefer that this not lead to budgets being set to 0.  OK, we can do that.  We don't need to do that in this case.

martinthomson:
I think that I've concluded that we probably want "impressions" as a new item for `Clear-Site-Data`.  That's not great, because we'd need to properly work through how that works in order to integrate, but I think that we should do it right.

[...]

PR #126, opened 2025-04-03
Some followup changes regarding epochs

martinthomson:
Looking at @csharrison here for a review.

I threw in a fix for #125 with this as a bonus.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/ppa/pull/126.html" title="Last updated on Apr 24, 2025, 11:10 PM UTC (99e6f04)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ppa/126/f63ab46...99e6f04.html" title="Last updated on Apr 24, 2025, 11:10 PM UTC (99e6f04)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -1067,6 +1067,11 @@ returning an [=epoch index=]:
     for the earliest [=epoch index=]
     that is supported for attribution.
 
+    <p class=note>
+    This value is not a constant [=epoch index=].
+    A value is chosen for each site
+    based on [=user agent=] preferences or configuration.
+
 1.  If the [=last browsing history clear=] is set,
     perform the following steps:
 
--- a/api.bs
+++ b/api.bs
@@ -847,7 +847,7 @@ for=PrivateAttribution>measureConversion()</a> method to find matching
 
 <p class=note>
 Though this API enables the storage of data by sites,
-this does not use a [=storage key=].
+the [=impression store=] does not use a [=storage key=].
 
 
 ### Contents ### {#impression-store-contents}
--- a/api.bs
+++ b/api.bs
@@ -959,6 +959,32 @@ is averaged across all the contributions from all [=users=].
 The <dfn>epoch start store</dfn> is a [=map=] keyed by [=site=]
 with values that are [=moments=].
 
+A <dfn local-lt=epoch>privacy budget epoch</dfn> (or [=epoch=])
+identifies a period of time.
+The length of an [=epoch=] is fixed at one week or 7 [=days=],
+where a <dfn>day</dfn> is defined as 86400 seconds.
+
+The exact time span for an [=epoch=] is different
+for each combination of [=user agent=] and [=site=].
+The start time for a given pair of [=user agent=] and [=site=]
+is chosen randomly by the [=user agent=] when an epoch is first needed--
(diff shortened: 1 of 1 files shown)


api.bs
```diff
@@ -1959,12 +1987,18 @@ to ensure that [=privacy budgets=] are not inadvertently exceeded.
...
 
-When clearing site data,
+When clearing site data at the request of a [=user=],
 but retaining browsing history,
 a [=user agent=] can [=map/set=] the value
 in the [=privacy budget store=] to 0 for all [=epochs=]
 for all affected [=sites=].
 
+When clearing site data at the request of a [=site=],
+through the use of the [:Clear-Site-Data:] header,
```
apasel422, line 1997: Do we need to address the points from https://github.com/w3c/ppa/issues/125#issuecomment-2772458616?

api.bs
```diff
@@ -1027,12 +1055,16 @@ about interactions with affected [=sites=]
...
+Resetting the budget is necessary to ensure
+that no information about the gap in browsing history
+is exposed.
 
 To <dfn>get the starting epoch for attribution</dfn>
-given [=site=] site:
+given [=site=] site,
+returning an [=epoch index=]:
 
 1.  Let |startEpoch| be a [=user agent=]-defined value
```
csharrison, line 1066: I am still confused by this. Let's say my earliest epoch index is 0. Now consider a site that first used PPA 5 years ago, so its `epoch start store[0]` is 5 years ago. This means I need to support attribution with impressions from 5 years ago right?

Similarly, setting this to a negative number doesn't make sense since there will never be impressions < 5 years old. Setting it to a positive number doesn't make much sense because then sites that just started using the API have to wait arbitrarily long to get attribution, etc.

It seems like `|startEpoch|` needs to be a function of the epoch start store (which is one of the reasons I was confused about the use of an indexing-based approach in the first place)

api.bs
martinthomson: I think that the discussion there suggests - to me at least - that we need to keep that issue open and do a little more work.  This is probably enough for now, given that.

api.bs
martinthomson: > This means I need to support attribution with impressions from 5 years ago right?

The oldest epoch the browser supports might be 3 weeks ago, or 4 months, or 10 years.  That decision is independent of the value of `epoch start store[site]` (that store is indexed by site).

> setting this to a negative number doesn't make sense since there will never be impressions < 5 years 

The start epoch might be negative because sites can save impressions without causing the value of `epoch start store[site]` to be initialized.  That only happens as a side effect of requesting a conversion.

`|startEpoch|` is probably going to be the result of calling `get current epoch` with a moment of `|now| - |x|` where `|x|` is set to a duration that the browser decides (three months, six weeks, whatever).  I didn't think it worthwhile specifying that, because it's an implementation choice.

csharrison: Sorry, I think my confusion is stemming from the fact that I assumed |startEpoch| to be a user-agent defined _constant_, maybe we can just update with a note to say that this index may depend on state related to |site|.

csharrison (approved):

Issue #138, opened 2025-04-09
Conversion options permit empty impression sites without apparent utility


apasel422:
[`PrivateAttributionConversionOptions.impressionSites`](https://w3c.github.io/ppa/#dom-privateattributionconversionoptions-impressionsites) both permits and defaults to the empty list.

But in the [`common matching logic`](https://w3c.github.io/ppa/#common-matching-logic) as currently specified, an empty list will cause no impressions to match:

> If options.impressionSites does not contain impression’s impression site, continue.

It might be beneficial to treat the empty list as "match any", since there's no reason to invoke `measureConversion` if you intend to match nothing.

Related to #134, https://github.com/w3c/ppa/issues/137#issuecomment-2790422726, #132.

eriktaubeneck:
I'm not very experienced with the syntax used in the spec, but is it possible for use to declare that a `sequence` must be non-empty?

To the point of "match any", I do worry about this creating a vector for fraud. Since these API calls are public, a malicious actor could identify sites which register conversions using an empty `impressionSites` sequence, and then register impressions targeting that conversion site to either claim more credit or simply exhaust the sites privacy budget.

While we cannot provide a reasonable default, I do think we should consider requiring with `impressionSites` or `intermediarySite` be set. (See #140 for more details on how we should think about `intermediarySite`.)

martinthomson:
My understanding is that dictionary fields when present are different than those that are absent and we can condition behavior on that.  It's a bit annoying though, as it is better to define things as `sequence<foo> bar = []` and avoid that question entirely.

martinthomson:
I think that we worked out how this fits together in our discussions in Paris.  If the set is empty, then any conversion site can be used.  Thanks to Arpana for coming up with the "multiple distributors for the same product" case for this. This is what is implemented in the current specification.

Issue #137, opened 2025-04-08
Inconsistent set vs scalar for impression-side "conversion site"


apasel422:
The [`PrivateAttributionImpressionOptions.conversionSite` field](https://w3c.github.io/ppa/#dom-privateattributionimpressionoptions-conversionsite) is a single string, but [the equivalent impression-store field](https://w3c.github.io/ppa/#impression-conversion-sites) is a set.

csharrison:
I vote to make them both sets. On the ARA side we've definitely seen use-cases where impressions can target multiple sites.

apasel422:
I agree with making them both sets, though we do need to think about whether the empty set should therefore mean "match anything" or "match nothing," the latter having no benefit IIUC.

eriktaubeneck:
> On the ARA side we've definitely seen use-cases where impressions can target multiple sites.

In these cases, is it reasonable to assume that there is a common `intermediarySite` which would be calling `measureConversion()`? 

If so, it may be simpler to have `saveImpression` take either a single `conversionSite` or a single `intermediarySite`.

martinthomson:
We fixed this one already in #139.

PR #139, opened 2025-04-09
Use consistent set semantics for impression/conversion sites

### LINKED ISSUE #134: Impression-side "conversion site" field is not used
Per https://w3c.github.io/ppa/#dom-privateattributionimpressionoptions-conversionsite:

> conversionSite, of type USVString
The site where conversions for this impression may occur, identified by its domain name. The measureConversion() method will only attribute to this impression when called by the indicated site.

But the measure conversion algorithm never actually uses this field. It seems like this should be checked in the [`common matching logic` algorithm](https://w3c.github.io/ppa/#common-matching-logic).

apasel422:
Fixes #134, #137, #138

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/ppa-api/pull/139.html" title="Last updated on Apr 9, 2025, 5:27 PM UTC (fb89d1e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ppa/139/eba7f55...apasel422:fb89d1e.html" title="Last updated on Apr 9, 2025, 5:27 PM UTC (fb89d1e)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -777,7 +777,7 @@ The arguments to <a method for=PrivateAttribution>measureConversion()</a> are as
   <dt><dfn>filterData</dfn></dt>
   <dd>Only [=impressions=] having a [=impression/filter data=] value matching this value will be eligible to match this [=conversion=]. If omitted, all [=impression/filter data=] values will match.</dd>
   <dt><dfn>impressionSites</dfn></dt>
-  <dd>A [=set=] of impression sites. Only [=impressions=] recorded where the [=impression site=] is in this set are eligible to match this [=conversion=].</dd>
+  <dd>A [=set=] of impression sites. Only [=impressions=] recorded where the [=impression site=] is in this set are eligible to match this [=conversion=]. If empty, any site will match.</dd>
   <dt><dfn>intermediarySites</dfn></dt>
   <dd>
     A [=set=] of sites which called the <a method for=PrivateAttribution>saveImpression()</a> API.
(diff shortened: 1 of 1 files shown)


csharrison (approved):

api.bs
```diff
@@ -1214,12 +1214,17 @@ To perform <dfn>common matching logic</dfn>, given
     1.  If |now| - |lookbackDays| is after |impression|'s [=impression/timestamp=],
         [=iteration/continue=].
 
+    1.  If |impression|'s [=impression/conversion sites=] [=set/is empty|is not empty=]
```
martinthomson, line 1217: Isn't it impossible to have an empty set?

api.bs
```diff
@@ -608,12 +608,12 @@ The arguments to <a method for=PrivateAttribution>saveImpression()</a> are as fo
...
+  <dt><dfn>conversionSites</dfn></dt>
   <dd>
-    The site where [=conversions=] for this impression may occur, identified by
-    its domain name. The <a method for=PrivateAttribution>measureConversion()</a>
-    method will only attribute to this impression when called by the indicated
-    site.
+    The sites where [=conversions=] for this impression may occur, identified by
+    their domain names. The <a method for=PrivateAttribution>measureConversion()</a>
+    method will only attribute to this impression when called by one of the indicated
+    sites. If empty, any site will match.
```
martinthomson, line 616: Do we really want this to be empty?  Is there any situation where it makes sense to avoid specifying this?

martinthomson (commented):
This seems fine, but I'm unclear on the value of not specifying where conversions occur.

api.bs
apasel422: No, this PR currently allows there to be an empty set of conversion sites, unless I made a mistake somewhere.

api.bs
apasel422: If you'd prefer, we could instead make the empty list an error.

CC @csharrison

martinthomson: How about we take this (which is unambiguously OK), then we can put the question about empty/not to the broader group.

PR #149, opened 2025-05-01
Fingerprinting risk of unconfigured browsers

### LINKED ISSUE #129: Getting aggregation service config to browsers
We need to write down our expectations about the availability of aggregation service configuration.

Our goal is that the measureConversion API calls cannot fail.  That means that browser need to have the configuration available always.  Either the config ships with the browser code, or there is some way to hide the fact that the browser is fetching that configuration.

martinthomson:
Closes #129.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/ppa/pull/149.html" title="Last updated on May 20, 2025, 1:32 AM UTC (8a0896b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ppa/149/b26b200...8a0896b.html" title="Last updated on May 20, 2025, 1:32 AM UTC (8a0896b)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -2073,12 +2073,12 @@ If the retrieval of configuration
 could delay the resolution of a call to <a for=PrivateAttribution>measureConversion()</a>,
 this creates a timing side channel.
 If a fake response is generated instead, differences between
-a response from a properly-configured [=user agent=]
+a response from a properly configured [=user agent=]
 and a fake response
 might be observable.
 This might be the use of key identifiers or
 other unencrypted metadata.
-A difference in ciphertext length might also reveal changes to what is enciphered
+A difference in ciphertext length might also reveal changes to what is encrypted
 or changes in choice of algorithm.
 
 Detecting stale or absent configuration might be used
--- a/api.bs
+++ b/api.bs
@@ -1952,7 +1952,7 @@ not to participate in attribution measurement.
 As discussed in [[#opt-out]], the user agent must provide
 a mechanism for the user to disable the Private Attribution API.
 
-To minimize the risk of fingerprinting,
+To minimize the risk of [=fingerprinting=],
 and to prevent discrimination
 against users who choose to disable the Private Attribution API,
 sites must not be able to detect that the API is disabled.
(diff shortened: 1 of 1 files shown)


api.bs
```diff
@@ -1967,6 +1967,32 @@ this difference cannot be detected
 by the site receiving the conversion report.
 
 
+## Unconfigured Browsers ## {#unconfigured}
+
+This API assumes that [=user agent=] instance
```
apasel422, line 1972:
```suggestion
This API assumes that [=user agent=] instances
```

api.bs
```diff
@@ -1967,6 +1967,32 @@ this difference cannot be detected
...
+prioritize the acquisition of [=aggregation service=] configuration
+in a way that is not observable to sites.
+This might be achieved by obtaining configuration at startup,
+before starting to load any content.
+
+If the necessary configuration cannot be obtained,
+or is obviously stale,
+an implementation can provide randomized responses
+that are indistinguishable from a real response
+when <a for=PrivateAttribution>measureConversion()</a> is invoked.
```
martinthomson, line 1993: This might not work for things like key identifiers.  We need to double check that for the different aggregation services.

martinthomson (commented):
Might be worth indicating what the side channel is: timing is one, the other is the metadata associated with the configuration, like key identifiers and whatnot.

api.bs
```diff
@@ -2058,6 +2058,46 @@ this difference cannot be detected
...
+If the retrieval of configuration
+could delay the resolution of a call to <a for=PrivateAttribution>measureConversion()</a>,
+this creates a timing side channel.
+If a fake response is generated instead, differences between
+a response from a properly-configured [=user agent=]
+and a fake response
+might be observable.
+This might be the use of key identifiers or
+other unencrypted metadata.
+A difference in ciphertext length might also reveal changes to what is enciphered
```
apasel422, line 2081:
```suggestion
A difference in ciphertext length might also reveal changes to what is encrypted
```

PR #158, opened 2025-05-15
Allow conversions to match a set of filter data values

### LINKED ISSUE #132: Consider allowing conversions to specify a set of integers to match filter data
Following from https://github.com/w3c/ppa/issues/58#issuecomment-2772991556, it might be useful to allow conversions to provide a set of values to match against filter data, rather than just one. If the given set is empty, all impressions match. (There is no reason to call `measureConversion` if you intend to match zero impressions.) Otherwise, an impression matches if the set contains its filter data.

This effectively allows any predicate to be implemented, assuming that the full set of impression-side filter data is remembered by the caller of `measureConversion`, making any subsequent improvements here mostly a form of compression, rather than new behavior.

This also maps nicely to the HTTP API via an inner list, if conversions can ever be registered using HTTP. We would permit responses like:

```http
Save-Conversion: ..., filter-data=12
```

```http
Save-Conversion: ..., filter-data=(12 13)
```
with the `filter-data` key's absence defaulting to `()`, meaning match anything.

apasel422:
When the set is empty, any value matches.

Fixes #132.

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -610,15 +610,6 @@ The arguments to <a method for=PrivateAttribution>saveImpression()</a> are as fo
     An optional piece of metadata associated with the impression. The value
     can be used to identify which impressions may receive attribution
     from a [=conversion=].
-    <p class=note>It is intentional that on the impression side this field
-    defaults to 0, while on
-    {{PrivateAttributionConversionOptions/filterData|the conversion side}} it is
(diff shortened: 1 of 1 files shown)


martinthomson (approved):
Perfect.  I assume you want to do the rename next.  This text already contains a naming change, I observe.

PR #161, opened 2025-05-16
Deal with privacy budget rounding

### LINKED ISSUE #77: Avoid using floating point arithmetic for privacy budget
This is a fairly minor issue, but it has been bugging me since reading the spec. We capture the value of `epsilon` in `PrivateAttributionConversionOptions` as a `double`, which implies (though it is not specified) that we will be doing accounting in `double` space.

Since we will be doing repeated addition / subtraction here, there is a chance for (small) errors to accumulate, meaning that our guarantee of `epsilon` may be (very) slightly off. My preference instead would be to have the interface be in `milliEpsilon`s and do accounting only with integers.[^1]

Note: an alternative would be to round `epsilon` to a fixed precision first, and then do error-free`epsilon` computations using rounded values. This has the benefit of enabling a possibly "cleaner" interface for developers, but is worse in my opinion because the rounding is not very transparent.

[^1]: I actually wish we had a rational type, but unfortunately it doesn't exist.

martinthomson:
This converts the underlying type from a CSS float (yes, we were citing CSS, not a numeric type) to a 32-bit integer.  I figure that an epsilon of 4000 is unnecessary, so this seemed like a nice way to manage it.

The algorithm is slightly different and I've tweaked some of the references to it.  Most of that is just referencing and cleanup.

Closes #77.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/ppa/pull/161.html" title="Last updated on May 20, 2025, 1:35 AM UTC (8dae34b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ppa/161/b01b4b4...8dae34b.html" title="Last updated on May 20, 2025, 1:35 AM UTC (8dae34b)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -1017,7 +1017,7 @@ and integer |globalSensitivity|:
     and return false.
 
 1.  [=map/set|Set=] the value of |key| in the [=privacy budget store=]
-    to |currentValue| - |deduction|
+    to |currentValue| − |deduction|
     and return true.
 
 
--- a/api.bs
+++ b/api.bs
@@ -973,6 +973,9 @@ in units [=microepsilons=].
 A <dfn>microepsilon</dfn> is one-millionth of the unitary epsilon value
 used in differential privacy [[DP]].
 
+<p class=note>  The choice of a 32 bit integer restricts the setting of epsilon to be less than 4295.
+This should be more than sufficient for implementations.
+
 A <dfn>privacy budget key</dfn> is a [=tuple=] consisting of the following items:
 
 <dl dfn-for="privacy budget key">
--- a/api.bs
+++ b/api.bs
@@ -1383,7 +1383,7 @@ set on a response requesting that the user agent invoke the
 <a method for=PrivateAttribution>saveImpression()</a> API.
 
 <pre class=example id=ex-save-impression-header>
-Save-Impression: conversion-site="advertiser.example", histogram-index=2, filter-data=12, lifetime-days=7
+Save-Impression: conversion-sites=("advertiser.example"), histogram-index=2, filter-data=12, lifetime-days=7
 </pre>
 
 The following keys are defined, corresponding to the members of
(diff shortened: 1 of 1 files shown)


api.bs
```diff
@@ -981,23 +985,37 @@ A <dfn>privacy budget key</dfn> is a [=tuple=] consisting of the following items
...
+    the budget to be exceeded unnecessarily after multiple invocations.
+    The privacy loss from the additional one-thousandth of an epsilon is trivial.
 
 1.  Let |currentValue| be the result of [=map/get|getting the value=] of |key|
     in the [=privacy budget store=].
 
-1.  If |currentValue| is less than or equal to 0, return false.
+1.  Let |deductionFp| be |epsilon| * |sensitivity| / |globalSensitivity|.
+
+1.  If |deductionFp| is negative or greater than 4294,
```
apasel422, line 1006: A note would be helpful to explain where `4294` is coming from. And since that value is also used below, maybe it should be pulled into a constant definition.

api.bs
```diff
@@ -981,23 +985,37 @@ A <dfn>privacy budget key</dfn> is a [=tuple=] consisting of the following items
...
+1.  Let |deduction| be |deductionFp| * 1000000, rounded towards positive Infinity.
 
-1.  [=map/set|Set=] the value of |key| in the [=privacy budget store=] to |newValue|.
+1.  If |deduction| is greater than |currentValue|,
+    [=map/set|set=] the value of |key| in the [=privacy budget store=] to 0
+    and return false.
 
-1.  Return whether |newValue| is greater than or equal to 0.
+1.  [=map/set|Set=] the value of |key| in the [=privacy budget store=]
+    to |currentValue| - |deduction|
```
apasel422, line 1017:
```suggestion
    to |currentValue| − |deduction|
```

apasel422 (changes requested):

api.bs
```diff
@@ -967,7 +967,11 @@ updated to refer to the [=privacy budget store=] as well.
 ### Privacy Budget Store ### {#s-privacy-budget-store}
 
 The <dfn>privacy budget store</dfn> is a [=map=] whose keys are
-[=privacy budget keys=] and whose values are [=floats=].
+[=privacy budget keys=] and whose values are [=32-bit unsigned integers=].
+These [=32-bit unsigned integer|integers=] store a value
+in units [=microepsilons=].
+A <dfn>microepsilon</dfn> is one-millionth of the unitary epsilon value
+used in differential privacy [[DP]].
```
csharrison, line 974:
```suggestion
used in differential privacy [[DP]].

<p class=note>  The choice of a 32 bit integer restricts the setting of epsilon to be less than 4295.
This should be more than sufficient for implementations.
```

csharrison (approved):

api.bs
michaelkleber: should that say "4095" instead?

csharrison: I don't think so, it is the result of `floor((2^32 - 1) / 1_000_000)`

michaelkleber: Oh got it; my bad for dividing by 1_048_576 instead

api.bs
```diff
@@ -967,7 +967,14 @@ updated to refer to the [=privacy budget store=] as well.
...
+
+<p class=note>  The choice of a 32 bit integer restricts the setting of epsilon to be less than 4295.
```
apasel422, line 976:
```suggestion
<p class=note>The choice of a 32-bit integer restricts the setting of epsilon to be less than 4295.
```

PR #163, opened 2025-05-16
Deal with maxValue and DAP proofs

### LINKED ISSUE #107: Limited valid values of maxValue
The [current (proposed) approach for summation in Prio](https://www.ietf.org/archive/id/draft-thomson-ppm-l1-bound-sum-00.html) uses an L1 bound that cannot take an arbitrary value.  Instead, the approach takes a shortcut and requires that the L1 bound is $2^n-1$ where $n\in\mathcal{N}$.  (We could adopt a different scheme, but that would require a lot larger submission size, because we'd need to provide additional witness values for the proof that Prio uses, I believe.)

This means that `maxValue` requires additional validation.

I don't know how we might best capture this in the API.  It seems like this is something that will vary based on the chosen ***aggregationService***, rather than the chosen attribution logic.  That's unfortunate, but it seems natural that a TEE would be better able to handle an arbitrary cap.

Could we instead apply the same restriction uniformly?

martinthomson:
This explains how the PrioL1BoundSum proofs are constructed from the maxValue that is provided.

Some cleanup of the surrounding text was included so that the DAP section has some structure.

Closes #107.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/ppa/pull/163.html" title="Last updated on May 19, 2025, 5:00 AM UTC (754a393)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ppa/163/b01b4b4...754a393.html" title="Last updated on May 19, 2025, 5:00 AM UTC (754a393)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -1482,7 +1482,7 @@ reports contain a distributed zero-knowledge proof
 that allows the nodes that participate in the MPC
 to confirm that the sum of the submitted histogram is less than a set value.
 [=Prio3L1BoundSum=] can only validate that the histogram sum
-is less than a power of two.
+is strictly less than a power of two.
 That is, proofs confirm that
 the sum is less than 2<sup><var>n</var></sup> for positive integer values of |n|.
 
@@ -1497,10 +1497,10 @@ Consequently, a malicious [=user agent=] might generate a report
 that contributes as much as double the value of {{PrivateAttributionConversionOptions/maxValue}}
 to the aggregate histogram.
 The proof ensures that there is no opportunity for excess contributions
-when {{PrivateAttributionConversionOptions/maxValue}} is equal to 2<sup><var>n</var></sup>-1;
+when {{PrivateAttributionConversionOptions/maxValue}} is equal to 2<sup><var>n</var></sup> − 1;
 the relative effect of any such attack can be reduced
 by setting {{PrivateAttributionConversionOptions/maxValue}}
-as close to 2<sup><var>n</var></sup>-1 as other constraints allow.
+as close to 2<sup><var>n</var></sup> − 1 as other constraints allow.
 
 
 ### DAP Extensions ### {#dap-extensions}
--- a/api.bs
+++ b/api.bs
@@ -1461,20 +1461,49 @@ This MPC configuration does not protect
 against the corruption of the outputs
 by either MPC operator.
 
+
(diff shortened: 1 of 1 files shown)


api.bs
```diff
@@ -1461,21 +1461,50 @@ This MPC configuration does not protect
...
 DAP also defines how an aggregate is obtained
 and what configuration is necessary
 for a user agent to obtain about the aggregation service.
 
+In using [=Prio3L1BoundSum=],
+reports contain a distributed zero-knowledge proof
+that allows the nodes that participate in the MPC
+to confirm that the sum of the submitted histogram is less than a set value.
+[=Prio3L1BoundSum=] can only validate that the histogram sum
+is less than a power of two.
```
csharrison, line 1485:
```suggestion
is strictly less than a power of two.
```

csharrison (approved):

api.bs
```diff
@@ -1461,21 +1461,50 @@ This MPC configuration does not protect
...
+that is the next power of two
+greater than {{PrivateAttributionConversionOptions/maxValue}}.
+This means that the [=aggregation service=] is unable to reject reports
+between {{PrivateAttributionConversionOptions/maxValue}}
+and the next power of two.
+Consequently, a malicious [=user agent=] might generate a report
+that contributes as much as double the value of {{PrivateAttributionConversionOptions/maxValue}}
+to the aggregate histogram.
+The proof ensures that there is no opportunity for excess contributions
+when {{PrivateAttributionConversionOptions/maxValue}} is equal to 2<sup><var>n</var></sup>-1;
```
apasel422, line 1500:
```suggestion
when {{PrivateAttributionConversionOptions/maxValue}} is equal to 2<sup><var>n</var></sup> − 1;
```

api.bs
```diff
@@ -1461,21 +1461,50 @@ This MPC configuration does not protect
...
+the relative effect of any such attack can be reduced
+by setting {{PrivateAttributionConversionOptions/maxValue}}
+as close to 2<sup><var>n</var></sup>-1 as other constraints allow.
```
apasel422, line 1503:
```suggestion
as close to 2<sup><var>n</var></sup> − 1 as other constraints allow.
```

apasel422 (changes requested):

apasel422 (approved):

PR #165, opened 2025-05-20
Require epsilon to be positive, rather than non-negative

apasel422:
Laplace(1/0) is not well-defined.


martinthomson (approved):
So, positive zero?  Maybe greater than 0 instead?

PR #198, opened 2025-06-04
Reject impressions with histogramIndex >= maximum histogram size

### LINKED ISSUE #197: Save impression method does not validate histogramIndex against maximum histogram size
IIUC it should probably be checked against [the same implementation-defined maximum used in `measureConversion`](https://w3c.github.io/ppa/#ref-for-dom-privateattributionconversionoptions-histogramsize%E2%91%A0) in order to avoid storing impressions that can *never* be used to fill a histogram ([Step 5 here](https://w3c.github.io/ppa/#fill-a-histogram-with-last-touch-attribution)).

apasel422:
Since they will never be able to fill a histogram.

Fixes #197

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/ppa-api/pull/198.html" title="Last updated on Jun 23, 2025, 11:58 AM UTC (6159fcd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/ppa/198/e6966b1...apasel422:6159fcd.html" title="Last updated on Jun 23, 2025, 11:58 AM UTC (6159fcd)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -1364,6 +1364,8 @@ To <dfn>validate {{AttributionConversionOptions}}</dfn> |options|:
     throw a {{RangeError}}.
 1.  If |options|.{{AttributionConversionOptions/histogramSize}}
     is 0 or greater than the [=implementation-defined=] <dfn>maximum histogram size</dfn>,
+    or is greater than the <dfn ignore>maximum aggregation-service histogram size</dfn>,
+    if any, for |options|.{{AttributionConversionOptions/aggregationService}},
     throw a {{RangeError}}.
 1.  Switch on the value of |options|.{{AttributionConversionOptions/logic}}:
     <dl class="switch">


martinthomson (commented):
It occurs to me that there is a risk that the limit is dictated by the aggregation service.  That's fine on the conversion side, because we have the details of the chosen aggregation service and the associated method available.  That's not going to be true on the impression-saving side.

This can probably only catch the case where the index exceeds the maximum across all aggregation services (or their associated processes).  WDYT?

csharrison:
I think that should be a separate limit and error. It is natural for the client to impose a different maximum which covers both impression and conversion.

martinthomson (changes requested):
Just flagging this as needing changes as we agreed.

PR #238, opened 2025-08-08
Miscellaneous simulator improvements

apasel422:


impl/src/backend.test.ts
```diff
@@ -0,0 +1,99 @@
...
+// TODO: Check that the distribution of credit arrays matches expectations.
+function runFairlyAllocateCreditTest(
+  tc: Readonly<FairlyAllocateCreditTestCase>,
+): void {
+  const sumCredit = tc.credit.reduce((a, b) => a + b, 0);
+  const normalizedFloatCredit = tc.credit.map(
+    (item) => (tc.value * item) / sumCredit,
+  );
+
+  const [rand, k] = tc.needsRand ? [Math.random, 10000] : [noRand, 1];
```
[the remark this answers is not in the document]
apasel422: Heh, indeed. That's the `TODO` above. Still thinking about the proper way to do so.

impl/src/backend.test.ts
```diff
@@ -0,0 +1,148 @@
...
+      tc.value,
+      `actual credit does not sum to value: ${actualCredit.join(", ")}`,
+    );
+  }
+
+  const intervals = getAllIntervals(
+    k,
+    tc.credit.map((c) => c / sumCredit),
+    0.0001,
+  );
```
csharrison, line 87:
```suggestion
  // Significance level for the overall test, where each subtest has
  // a null hypothesis that its correct mean is precisely the passed in
  // normalized float credit. This means the test should flake with
  // probability no more than `alphaTotal`.
  const alphaTotal = .0001
  const intervals = getAllIntervals(
    k,
    normalizedFloatCredit,
    alphaTotal
  );
```

apasel422: Maybe the variables aren't named properly, but currently `normalizedFloatCredit` has a `value` factor, whereas, IIUC, `getAllIntervals` must operate on credit values in `[0, 1]`.

csharrison: Oh you're right. I think it is expecting `normalizedFloatCredit - floor(normalizedFloatCredit)` i.e. the fractional part.

impl/src/backend.test.ts
```diff
@@ -0,0 +1,152 @@
...
+      assert.ok(
+        diff < 1,
+        `credit error >= 1: actual=${actual}, normalized=${normalized}`,
+      );
+
+      totals[j]! += actual / tc.value;
+    }
+
+    assert.equal(
+      // TODO: replace with precise sum
```
martinthomson, line 80: Given the number and range of values we're talking about, I don't know that preciseSum will make a big difference here.

impl/src/backend.test.ts
```diff
@@ -0,0 +1,152 @@
...
+function normalPpf(q: number, mean: number, stdDev: number): number {
+  return (
+    mean +
+    stdDev * Math.sqrt(2) * simplestatistics.inverseErrorFunction(2 * q - 1)
+  );
+}
+
+function getIntervalApprox(n: number, p: number, alpha: number): Interval {
+  // Approximates a binomial distribution with a normal distribution which is a bit
+  // simpler as it is symmetric.
```
martinthomson, line 30: I would assert that `n` is large enough for this approximation to be useful.

impl/src/backend.test.ts
```diff
@@ -0,0 +1,152 @@
...
+function noRand(): number {
+  throw new Error("no rand expected");
+}
+
+type Interval = [min: number, max: number];
+
+// https://en.wikipedia.org/wiki/Normal_distribution#Quantile_function
+function normalPpf(q: number, mean: number, stdDev: number): number {
+  return (
+    mean +
```
martinthomson, line 23: Why are you taking `mean` in here, when you only ever pass 0?

impl/src/backend.test.ts
```diff
@@ -0,0 +1,152 @@
...
+  needsRand?: boolean;
+}
+
+function noRand(): number {
+  throw new Error("no rand expected");
+}
+
+type Interval = [min: number, max: number];
+
+// https://en.wikipedia.org/wiki/Normal_distribution#Quantile_function
```
martinthomson, line 20: https://en.wikipedia.org/wiki/Probit seems clearer to me.  It defines $\mathbb{\Phi}^{-1}$ further down.

impl/src/backend.test.ts
```diff
@@ -0,0 +1,152 @@
...
+  return [mean - diff, mean + diff];
+}
+
+function getAllIntervals(
+  n: number,
+  creditFractions: readonly number[],
+  alphaTotal: number,
+): Interval[] {
+  // When testing multiple hypotheses, dividing `alphaTotal` by the dimension
+  // for each individual test is sometimes called the "bonferroni correction".
```
martinthomson, line 43:
```suggestion
  // We are testing one hypothesis per dimension, so divide `alphaTotal` by 
  // the number of dimensions: https://en.wikipedia.org/wiki/Bonferroni_correction
```

impl/src/backend.test.ts
```diff
@@ -0,0 +1,152 @@
...
+  const intervals = getAllIntervals(
+    k,
+    tc.credit.map((c) => c / sumCredit),
+    0.0001,
+  );
+  for (const [j, total] of totals.entries()) {
+    const [min, max] = intervals[j]!;
+    assert.ok(
+      total >= min && total <= max,
+      `total for credit[${j}] ${total} not in confidence interval [${min}, ${max}]`,
```
martinthomson, line 96:
```suggestion
      `total for credit[${j}] ${total} not in ${1 - alpha} confidence interval [${min}, ${max}]`,
```

impl/src/backend.test.ts
martinthomson: So that, and I would go further and avoid using the interval test when `k == 1`:

```suggestion
  const alpha = 0.00001; // Probability of test failing at random.
  const intervals = (k > 1) ? getAllIntervals(
    k,
    normalizedFloatCredit.map((c) => c - Math.floor(c)),
    alpha,
  ) : normalizedFloatCredit.map((c) => [c, c]);
```

I would reduce the value of `k` and increase `alpha` given this arrangement.  No need to run 100k times when `alpha` is what governs the probability of failure: increase that by a factor of 10 instead.

impl/src/backend.test.ts
```diff
@@ -0,0 +1,152 @@
...
+function runFairlyAllocateCreditTest(
+  tc: Readonly<FairlyAllocateCreditTestCase>,
+): void {
+  // TODO: replace with precise sum
+  const sumCredit = tc.credit.reduce((a, b) => a + b, 0);
+  const normalizedFloatCredit = tc.credit.map(
+    (item) => (tc.value * item) / sumCredit,
+  );
+
+  const [rand, k] = tc.needsRand ? [Math.random, 100000] : [noRand, 1];
```
martinthomson, line 57:
```suggestion
  const [rand, k] = tc.needsRand ? [Math.random, 1000] : [noRand, 1];
```

No point thrashing TOO hard if you have a solid statistical test.

impl/src/backend.test.ts
apasel422: Sure, but there's no reason not to use it if it's available.

Issue #212, opened 2025-06-24
Clarify what we mean by `sensitivity`


csharrison:
Branched from discussion https://github.com/w3c/ppa/pull/211.

In the spec, we use `sensitivity` and `globalSensitivity` to refer to the maximum L1 norm for any vector contribution, for a particular conversion and across all conversions respectively. However, this is _not_ the same as the sensitivity induced by a DP neighbor e.g. in [Cookie Monster](https://arxiv.org/pdf/2405.16719), since e.g. removing a set of events contributing $\Delta$ to some histogram can cause a cascade to some other epoch, causing a $2 \Delta$ difference between the histogram outputs on neighboring DBs.

As such, we'd need to add (for instance) $Lap(2 \Delta / \epsilon)$ to satisfy $\epsilon$-DP under the Cookie Monster adjacency notion. We should clarify this point in the spec, possibly by disambiguating what we mean by "sensitivity".

cc @tholop

csharrison:
Here's an example to make sure we're on the same page (cc @martinthomson ). Consider the following events:

1. Impression 1 in epoch 1
2. Impression 2 in epoch 2
3. Last touch conversion with value 1

For impression-level keys this emits the non-noisy histogram (0, 1). Under a neighboring dataset where impression 2 is removed, this emits (1, 0), which has 2x the L1 sensitivity of the raw contribution vector. To hide this particular change from an adversary requires invoking the Laplace mechanism with $Lap(2 / \epsilon)$.

martinthomson:
I had been operating on the model where the sensitivity is based on the difference between a user participating or not.  In that scenario, your (1,0) or (0,1) is relative to the (0,0) of non-participation. 

I can definitely see how you might conclude that sensitivity is doubled though.  It's entirely possible that the simplistic participation/not isn't the right framing.  I don't have the necessary tools to be able to decide which is correct.

csharrison:
> I had been operating on the model where the sensitivity is based on the difference between a user participating or not. In that scenario, your (1,0) or (0,1) is relative to the (0,0) of non-participation.

I don't think that's correct. The "unit of participation" should be the privacy unit. In the example I gave, we simulate non-participation by removing just events in a single epoch (e.g. epoch 2). The challenge with attribution is that this causes a cascading effect and histogram mass gets redistributed.

There isn't a natural way to move to an adjacent db that induces (0, 0) without a broader privacy unit that removes both impression 1 and impression 2.

PR #239, opened 2025-08-11
Add single epoch optimization

### LINKED ISSUE #78: Spec the single epoch optimization

csharrison:
This fixes #78 and partially addresses #212 . The algorithm uses the configured lookback window to determine automatically if the attribution is only considering impressions from the current epoch. If so, budget deduction is altered by considering the report's sensitivity to be `|l1Norm|` rather than `2 * |value|`.

In both cases, we assume that the noise scale for Laplace noise is `lambda = |maxValue| / |epsilon|.

Note: An alternative considered was to add a new option to the API to query epochs and apply the optimization if only a single epoch was chosen, but _using_ that API seems difficult, and at a minimum requires exposing the epoch start map to all conversion sites to use it effectively. Additionally, it adds API surface bloat.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/pull/239.html" title="Last updated on Aug 15, 2025, 6:45 PM UTC (a3d8244)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/239/12c9995...a3d8244.html" title="Last updated on Aug 15, 2025, 6:45 PM UTC (a3d8244)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -1143,7 +1143,7 @@ given a [=privacy budget key=] |key|,
 [[WEBIDL#idl-double|double]] |epsilon|,
 integer |value|,
 integer |maxValue|, and
-integer |l1Norm|:
+integer |attributedValueForSingleEpochOpt|:
 
 1.  If the [=privacy budget store=] does not [=map/contain=] |key|, [=map/set=]
     its value of |key| to be a [=user agent=]-defined value,
@@ -1157,13 +1157,13 @@ integer |l1Norm|:
 1.  Let |currentValue| be the result of [=map/get|getting the value=] of |key|
     in the [=privacy budget store=].
 
-1.  Let |singleEpochQuery| be true if |l1Norm| is non-null, false otherwise.
+1.  Let |singleEpochQuery| be true if |attributedValueForSingleEpochOpt| is non-null, false otherwise.
 
-1.  Let |reportGlobalSensitivity| be |l1Norm| if |singleEpochQuery| is true, 2 * |value| otherwise.
+1.  Let |halfReportGlobalSensitivity| be |attributedValueForSingleEpochOpt|/2 if |singleEpochQuery| is true, |value| otherwise.
 
-1.  Let |noiseScale| be |maxValue| / |epsilon|.
+1.  Let |noiseScale| be 2 * |maxValue| / |epsilon|.
 
-1.  Let |deductionFp| be |reportGlobalSensitivity| / |noiseScale|.
+1.  Let |deductionFp| be |halfReportGlobalSensitivity| / |noiseScale|.
 
 <p class=note>Single epoch attributions do not cause any cascading effects across epochs,
 so they do not need to consume double the budget. The deduction here is assuming Laplace


api.bs
```diff
@@ -1598,21 +1609,33 @@ To <dfn>do attribution and fill a histogram</dfn>, given
...
+        with |options|, |topLevelSite|, |intermediarySite|, |currentEpoch|, and |now|.
+
+1.  If |singleEpoch| is false:
 
-    1.  If |impressions| [=set/is empty|is not empty=]:
+    1.  For each |epoch| from |startEpoch| to |currentEpoch|, inclusive:
 
-        1.  Let |key| be a [=privacy budget key=] whose items are |epoch| and |topLevelSite|.
+        1.  Let |impressions| be the result of invoking [=common matching logic=]
+            with |options|, |topLevelSite|, |intermediarySite|, |epoch|, and |now|.
```
martinthomson, line 1626: So we have this problem, whereby we let the multi-epoch query avoid deductions unless it selects an impression from an epoch.  But the single epoch query does not.  I see that this is how it is implemented in psdlib, so I'm not contesting the conclusion, but it's annoying.

If a multi-epoch query only selects impressions from a single epoch, we deduct `2 * value * epsilon / maxValue`.  That's mostly OK, but that factor of 2 is a real challenge.  It is not going to be obvious to people using this API that the epsilon value in their browser is only half of what they get for most queries.  That is, the browser might set an epsilon budget of 4, but the site can only reasonably make queries with epsilon = 2 within that budget, unless they are careful to stay within a single epoch.  Given the length of our epoch - and its random start offset - that will rarely be useful to them, so I don't expect it to happen.

martinthomson (approved):

api.bs
csharrison: > So we have this problem, whereby we let the multi-epoch query avoid deductions unless it selects an impression from an epoch. But the single epoch query does not. I see that this is how it is implemented in psdlib, so I'm not contesting the conclusion, but it's annoying.

That is not the case. In the single epoch case we:

1. Return early if `matchedImpressions` is empty, deducting no budget (before invoking attribution logic)
2. If there are matched impressings, we deduct proportional to the actual realized L1 norm of the histogram, so even if we removed step (1) we would still deduct nothing.

> If a multi-epoch query only selects impressions from a single epoch, we deduct 2 * value * epsilon / maxValue. That's mostly OK, but that factor of 2 is a real challenge. It is not going to be obvious to people using this API that the epsilon value in their browser is only half of what they get for most queries. That is, the browser might set an epsilon budget of 4, but the site can only reasonably make queries with epsilon = 2 within that budget, unless they are careful to stay within a single epoch. Given the length of our epoch - and its random start offset - that will rarely be useful to them, so I don't expect it to happen.

Two things: 

a. Without the single-budget optimization, we still need the factor of 2 increase, per some of the discussion in #212. This PR only makes this more apparent (since in the existing spec, to achieve the privacy guarantees the noise factor needs to be `2 * |maxValue| / |epsilon|`.

b. I actually think there are plenty of use-cases where you will hit the single-budget opt. For instance, I think it is relatively common to have O(day) lookback windows for view throughs. In those cases, I think it is perfectly fine to "silently" optimize and deduct less than the expected budget.

csharrison:
Thanks folks, I'll wait for review by @alexanderknop before landing.

api.bs
```diff
@@ -1141,8 +1141,9 @@ A <dfn>privacy budget key</dfn> is a [=tuple=] consisting of the following items
 To <dfn>deduct privacy budget</dfn>
 given a [=privacy budget key=] |key|,
 [[WEBIDL#idl-double|double]] |epsilon|,
-integer |sensitivity|,
-and integer |globalSensitivity|:
+integer |value|,
+integer |maxValue|, and
```
alexanderknop, line 1145: I would suggest renaming `maxValue` into scale or `scalingCoefficient`

api.bs
```diff
@@ -1141,8 +1141,9 @@ A <dfn>privacy budget key</dfn> is a [=tuple=] consisting of the following items
...
+integer |l1Norm|:
```
alexanderknop, line 1146: I would rename l1Norm into attributedValue or actuallyAttributedValue

api.bs
```diff
@@ -1156,7 +1157,17 @@ and integer |globalSensitivity|:
 1.  Let |currentValue| be the result of [=map/get|getting the value=] of |key|
     in the [=privacy budget store=].
 
-1.  Let |deductionFp| be |epsilon| * |sensitivity| / |globalSensitivity|.
+1.  Let |singleEpochQuery| be true if |l1Norm| is non-null, false otherwise.
+
+1.  Let |reportGlobalSensitivity| be |l1Norm| if |singleEpochQuery| is true, 2 * |value| otherwise.
+
+1.  Let |noiseScale| be |maxValue| / |epsilon|.
```
alexanderknop, line 1164: For me, scaling the noise instead of the budget would make the API cleaner. 
In this case, I would rename reportGlobalSensitivity into halfOfReportGlobalSensitivity (it is a long name, but very self-explanatory).

alexanderknop (approved):

alexanderknop:
The algorithms loogs great to me so other than renaming variables for clarity, I think this is a great!

api.bs
csharrison: Can you elaborate a bit more on what you're looking for? Are you suggesting to 

1. Make the noise scale always be `2 * |maxValue| / |epsilon|`
2. Change the API surface so we let the agg service add less noise in this case

I think (1) is fine, but then I don't think that we should change the name of |maxValue| to |scaleCoefficient| unless we fold the factor of 2 into it.

api.bs
csharrison: Not addressing this now, but I can in a follow-up

PR #242, opened 2025-08-12
Pin python version for Check API workflow

csharrison:


martinthomson (approved):
Yeah, this is probably better yet.

PR #271, opened 2025-09-04
Add text on DP formal analysis and its assumptions

bmcase:
With @roxanageambasu and @tholop we have put together some text for the spec that states what assumptions are used in the system model in which the formal DP analysis is done. 

The referenced paper is already available, but a few claims in the PR are not yet reflected there. We will update the arXiv version to align with the PR shortly. In the meantime, we confirm that the statements are accurate and supported by analyses we already have internally.

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -2338,35 +2338,46 @@ for a number of reasons:
     Without allocating [=privacy budget=] for new data,
     sites could exhaust their budget forever.
 
-We operate the system with [=site=] in the [=privacy unit=], but theoretically proving
-DP guarantees for this privacy unit requires additional assumptions.
-Since different sites are interacting with the same user over time there is not
-complete independence between privacy units maintained for the same user which is necessary
-for proving the formal DP guarantees. These assumptions capture what would need to be assumed
-for these privacy units to be independent and under which the DP guarantees can be proven rigorously.
-* Assumption 1: Siloed adaptive data generation. The data stream of the user's impressions
-and conversions able to be queried by one site is generated independently from the past query
-results of other sites.
-* Assumption 2: No leakage through safety limits. The queries from one site do not impact
-which reports get filtered for any other site.
+### Formal Analysis of Privacy Properties and Their Limitations ### {#formal-analysis}
+The paper [[PPA-DP-2]] provides formal analysis of the mathematical privacy guarantees
+afforded by <dfn> per-site budgets </dfn> and by <dfn> safety limits </dfn>. Per-site
+budgets include [=site=] in the [=privacy unit=], whereas safety limits exclude it
+thereby enforcing a global individual DP guarantee.
 
-Both of these assumptions are technically not met by the system. By stating them here we
-are saying that the API does not protect against privacy leakages that result from them.
+The analysis shows that <dfn> per-site individual DP guarantees </dfn> hold under a restricted system
+ model that makes two assumptions, which may not always be satisfied in practice:
 
-Assumption 1 is a consequence of multiple sites interacting with the same user over time and
-changing the ads they show to users based on their use of this system. For example, one advertiser
-might learn how to make one of their ads more effective which results in the user converting on
-their site rather than on a different site. This impacts the data the other site has available to
-query and introduces across site adaptive dependency in how data is generated.
+1.  <dfn> No cross-site adaptivity in data generation. </dfn> A site’s queryable data stream (impressions
+and conversions) must be generated independently of past DP query results from other sites.
+2.  <dfn> No leakage through cross-site shared limits. </dfn> Queries from one site must not affect which
+reports are emitted to others.
 
-Assumption 2 is a consequence of having safety limits that span across multiple sites.
-If the safety limits are never triggered there is no change in the data available to
-be queried across all sites covered by the safety limit. But if as a result of some
-sites’ queries causing the safety limit to be reached and start filtering out reports
-sent to other sites, there is now a dependence introduced across the data queried under
- the separate privacy units.
+Assumption 1 is necessary because the system involves multiple sites that could interact
+with the same user over time and change the ads they show to the user, or impact the
+conversions the user has, based on each other’s DP measurements. For example, if one advertiser
+learns, from DP measurements, to make an ad more effective, a user may convert on their site
+rather than a competitor’s. In this case, the first site’s DP outputs -- counted only against
+its own per-site budget -- alter the data (or absence of data) visible to the competitor, yet
+this impact is not reflected in the competitor’s per-site budget. When Assumption 1 is violated,
+the analysis shows that per-site guarantees cannot be achieved.
 
-See [cite paper] for more details on the formal DP proofs and in depth discussion of these assumptions.
+Assumption 2 is necessary when we have shared limits that span multiple sites. An example of
+such shared limits are the global safety limits that aim to provide a global DP guarantee.
+If queries from some sites cause a shared limit to be reached, reports to other sites may be
+filtered, creating dependencies across separate per-site privacy units and affecting the validity
+of the per-site guarantees. Thus, care must be taken when introducing any new shared limit, such
+as cross-site rate limiters on privacy loss. If only Assumption 2 is violated, it is unknown whether
+per-site guarantees can still be preserved, for example via special designs of the shared limits.
+
+These results suggest that per-site protections should be regarded as theoretically grounded approximations
+of an ideal per-site individual DP guarantee that can be established only under certain assumptions.
+The extent to which privacy protection from per-site budgets may be impacted in practice remains unknown.
+
+By contrast, the analysis shows that <dfn> safety limits </dfn> -- which operate at global level,
+excluding [=site=]
+from the [=privacy unit=] -- can be implemented to deliver <dfn> sound global individual DP guarantees </dfn>
+regardless of whether either assumption is satisfied.
+
 
 ### Browser Instances ### {#dp-instance}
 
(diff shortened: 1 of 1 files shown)


csharrison:
Hey Ben, we are making some changes to the spec and assumptions on the noise addition in https://github.com/w3c/attribution/pull/264. Can you make sure this PR aligns with that one?

api.bs
```diff
@@ -2338,6 +2338,45 @@ for a number of reasons:
...
+reports are emitted to others.
+
+Assumption 1 is necessary because the system involves multiple sites that could interact
+with the same user over time and change the ads they show to the user, or impact the
+conversions the user has, based on each other’s DP measurements. For example, if one advertiser
+learns, from DP measurements, to make an ad more effective, a user may convert on their site
+rather than a competitor’s. In this case, the first site’s DP outputs -- counted only against
+its own per-site budget -- alter the data (or absence of data) visible to the competitor, yet
+this impact is not reflected in the competitor’s per-site budget. When Assumption 1 is violated,
+the analysis shows that per-site guarantees cannot be achieved.
```
martinthomson, line 2362: This is part of the assumption, but I think that the main challenge here is different.  Sites might gain an understanding that a particular visitor to each site in a set is the same person (due to federated login, same email address, or anything including navigation tracking which we can't stop).  AND THEN they decide to pool their per-site budgets to use the API to extract more information about that person.  In that case, we have no defense from the per-site budget.  Sites are only limited by their ability to link activity across sites (which is too easy, as noted) and then the global budget.

So we should acknowledge that limitation as well as the more theoretical one here.

api.bs
```diff
@@ -2338,6 +2338,45 @@ for a number of reasons:
...
+afforded by *per-site budgets* and by *safety limits*. Per-site
+budgets include [=site=] in the [=privacy unit=], whereas safety limits exclude it
+thereby enforcing a global individual DP guarantee.
+
+The analysis shows that *per-site individual DP guarantees* hold under a restricted system
+ model that makes two assumptions, which may not always be satisfied in practice:
+
+1.  *No cross-site adaptivity in data generation.* A site’s queryable data stream (impressions
+and conversions) must be generated independently of past DP query results from other sites.
+2.  *No leakage through cross-site shared limits.* Queries from one site must not affect which
```
martinthomson, line 2352:
```suggestion

1.  *No leakage through cross-site shared limits.* Queries from one site must not affect which
```

House style is that lists are all ones.

api.bs
```diff
@@ -2338,6 +2338,45 @@ for a number of reasons:
...
     sites could exhaust their budget forever.
 
+### Formal Analysis of Privacy Properties and Their Limitations ### {#formal-analysis}
+The paper [[PPA-DP-2]] provides formal analysis of the mathematical privacy guarantees
+afforded by *per-site budgets* and by *safety limits*. Per-site
+budgets include [=site=] in the [=privacy unit=], whereas safety limits exclude it
+thereby enforcing a global individual DP guarantee.
+
+The analysis shows that *per-site individual DP guarantees* hold under a restricted system
+ model that makes two assumptions, which may not always be satisfied in practice:
```
martinthomson, line 2348:
```suggestion
model that makes two assumptions, which may not always be satisfied in practice:
```

api.bs
```diff
@@ -2338,6 +2338,45 @@ for a number of reasons:
     Without allocating [=privacy budget=] for new data,
     sites could exhaust their budget forever.
 
+### Formal Analysis of Privacy Properties and Their Limitations ### {#formal-analysis}
```
martinthomson, line 2341: I think that you want to mention both papers in the introduction to this, not just the big bird one.  Point out that the first establishes the basic system that this document relies on and the second expands that to be a more comprehensive analysis of the system as a whole and the overall privacy guarantees.

This is a timely reminder that we need to implement safety limits in the spec.

api.bs
```diff
@@ -2338,6 +2338,45 @@ for a number of reasons:
...
+The paper [[PPA-DP-2]] provides formal analysis of the mathematical privacy guarantees
+afforded by *per-site budgets* and by *safety limits*. Per-site
```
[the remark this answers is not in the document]
bmcase: added

api.bs
bmcase: @martinthomson  I added a paragraph at the end of this section to capture this additional challenge an adversary is faced with.  Let me know if you think that still needs any adjusting. 

_In addition to facing safety limits discussed above, an attacker using multiple colluding sites to gain
more DP budget about users also face the practical limitation of being able to link a user across sites.
This is limitation does not itself provide a theoretical DP benefit but does impose a significant
challenge to the attacker when the user agent has made it difficult to link users across sites._

bmcase:
Updated the PR to address all the points of feedback above, including a summary of both papers and paragraph to address the challenge of cross-site linking.   Cleaned up the markdown format and bib such that all checks are passing now. 

We previously [discussed](https://github.com/w3c/patwg/blob/main/meetings/2025/09-telecons/09-09-minutes.md#httpsgithubcomw3cattributionpull271) this PR in a PATWG meeting and all were supportive of it.  

I think we should be ready to merge this PR now. cc @martinthomson , @csharrison , @apasel422 , @andyleiserson

api.bs
```diff
@@ -2344,6 +2344,53 @@ for a number of reasons:
...
 
+### Formal Analysis of Privacy Properties and Their Limitations ### {#formal-analysis}
+ The formal privacy analysis is based on two papers. The first [[PPA-DP]] establishes the theory
+for on-device Individual DP accounting. The second [[PPA-DP-2]] provides formal analysis of
+the mathematical privacy guarantees afforded by *per-site budgets* and by *safety limits*
+(Section 3 is about per-site guarantees and Section 3.4 specifies the assumptions under
+which they hold). Per-site budgets include [=site=] in the [=privacy unit=], whereas safety
+limits exclude it thereby enforcing a global individual DP guarantee.  In Attribution Level 1
+it is conversion sites that have per-site budgets tracked.
+
```
```suggestion
```

api.bs
```diff
@@ -2344,6 +2344,53 @@ for a number of reasons:
...
+
+The analysis shows that *per-site individual DP guarantees* hold under a restricted system
```
```suggestion
The analysis in [[???]] shows that *per-site individual DP guarantees* hold under a restricted system
```

api.bs
```diff
@@ -2344,6 +2344,53 @@ for a number of reasons:
...
+model that makes two assumptions, which may not always be satisfied in practice:
+
+1.  *No cross-site adaptivity in data generation.* A site's queryable data stream (impressions
+    and conversions) must be generated independently of past DP query results from other sites.
+1.  *No leakage through cross-site shared limits.* Queries from one site must not affect which
+    reports are emitted to others.
+
+Assumption 1 is necessary because the system involves multiple sites that could interact
```
```suggestion
The assumption that sites cannot adapt their queries is necessary
because the system involves multiple sites that could interact
```

api.bs
```diff
@@ -2344,6 +2344,53 @@ for a number of reasons:
...
+with the same user over time and change the ads they show to the user, or impact the
+conversions the user has, based on each other’s DP measurements. For example, if one advertiser
+learns, from DP measurements, to make an ad more effective, a user may convert on their site
```
```suggestion
learns generally-applicable information that helps them make their ads more effective,
that will make it more likely that their ads are attributed for conversions,
as opposed to a competitor.
```

Here, "DP measurements" refers to `measureConversion()` as well.  Watch out for that.

api.bs
```diff
@@ -2344,6 +2344,53 @@ for a number of reasons:
...
+rather than a competitor’s. In this case, the first site’s DP outputs -- counted only against
+its own per-site budget -- alter the data (or absence of data) visible to the competitor, yet
+this impact is not reflected in the competitor’s per-site budget. When Assumption 1 is violated,
+the analysis shows that per-site guarantees cannot be achieved.
+
+Assumption 2 is necessary when we have shared limits that span multiple sites. An example of
```
```suggestion
An assumption that sites are unable to coordinate their use of the API is necessary
when we have shared limits that span multiple sites. An example of
```

api.bs
```diff
@@ -2344,6 +2344,53 @@ for a number of reasons:
...
+such shared limits are the global safety limits that aim to provide a global DP guarantee.
+If queries from some sites cause a shared limit to be reached, reports to other sites may be
```
martinthomson: Again, queries... If you want to use that word, it might be necessary to explain it up front.

api.bs
```diff
@@ -2344,6 +2344,53 @@ for a number of reasons:
...
+The extent to which privacy protection from per-site budgets may be impacted in practice remains unknown.
+
+By contrast, the analysis shows that *safety limits* -- which operate at global level,
+excluding [=site=] from the [=privacy unit=] -- can be implemented to deliver *sound global individual
+DP guarantees* regardless of whether either assumption is satisfied.
+
+In addition to facing safety limits discussed above, an attacker using multiple colluding sites to gain
+more DP budget about users also face the practical limitation of being able to link a user across sites.
+This is limitation does not itself provide a theoretical DP benefit but does impose a significant
+challenge in practice to the attacker when the user agent has made such cross-site linking difficult.
```
```suggestion
An attacker that is able to use other information to link the activity of a user across multiple sites
can use the DP budgets of those sites to overcome the constraints of the per-site budgets.
There are many features in the web platform that allow this capability,
so this is a very plausible attack on the privacy design.
The privacy analysis does not consider this attack within its threat model,
relying exclusively on [=global safety limits=].
Any implementation needs to consider this limitation when selecting DP parameters.
```

api.bs
```diff
@@ -2344,6 +2344,53 @@ for a number of reasons:
...
+as cross-site rate limiters on privacy loss. If only Assumption 2 is violated, it is unknown whether
+per-site guarantees can still be preserved, for example via special designs of the shared limits.
+
+These results suggest that per-site protections should be regarded as theoretically grounded approximations
+of an ideal per-site individual DP guarantee that can be established only under certain assumptions.
+The extent to which privacy protection from per-site budgets may be impacted in practice remains unknown.
+
+By contrast, the analysis shows that *safety limits* -- which operate at global level,
+excluding [=site=] from the [=privacy unit=] -- can be implemented to deliver *sound global individual
+DP guarantees* regardless of whether either assumption is satisfied.
```
martinthomson: Consider defining this as `[=global safety limits=]`, per above.


```suggestion
The analysis shows that [=global safety limits=] --
which do not have a [=site=]-specific [=privacy unit=] --
deliver sound individual DP guarantees
regardless without relying on either of these assumptions.
```

api.bs
bmcase: What I'm trying to say here is just that if you want to have shared limits you have to make Assumption 2 for the per-site budgets to hold.

api.bs
bmcase: This a case where we are talking about aggregate results from many devices that you get back from the aggregation service. 

I feel we need some sort of term for this in the spec as for what to call the final aggregate DP results that go back to the advertiser.   Maybe "query" is too general but something like  "DP attribution results" or "DP measurements" should be clear and maybe we need to define that somewhere in the intro.

PR #276, opened 2025-09-10
Remove single-value AttributionLogic enum and inline credit field

### LINKED ISSUE #275: Consider simplifying AttributionConversionOptions for initial iteration of API
Since `last-n-touch` is the only supported `AttributionLogic` right now, and since no one has proposed an additional declarative attribution model (compare to #204), we may want to consider replacing

```webidl
partial dictionary AttributionConversionOptions {
  AttributionLogic logic = "last-n-touch";
  AttributionLogicOptions logicOptions;
};
```
with
```webidl
partial dictionary AttributionConversionOptions {
  sequence<double> credit;
};
```
for simplicity. An `AttributionLogic` enum could always be added in the future, defaulting to `last-n-touch` for backwards compatibility.

This would also slightly simplify #31.

apasel422:
Fixes #275

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -357,10 +357,10 @@ The histogram created by the [=conversion report=] is constructed as follows:
     or the [=privacy budget=] for the site is exhausted,
     a histogram consisting entirely of zeros (0) is constructed.
 
-*   If one or more matching impressions is found, the browser runs the attribution
-    logic (default last-n-touch) to select the most recent impression. The provided conversion
+*   If one or more matching impressions is found, the browser runs the last-n-touch
+    attribution logic to select the relevant impression(s). The provided conversion
     value is added to a histogram at the bucket that was specified at the time of the
-    attributed impression. All other buckets are set to zero.
+    attributed impression(s). All other buckets are set to zero.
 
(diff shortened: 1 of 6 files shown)


api.bs
```diff
@@ -1558,8 +1543,7 @@ The <dfn method for=Attribution>measureConversion(|options|)</dfn> method steps
 <dfn>Match Values</dfn>: A [=set=] of [=32-bit unsigned integers=].
 <dfn>Impression Sites</dfn>: A [=set=] of [=sites=].
 <dfn>Impression Callers</dfn>: A [=set=] of [=sites=].
-<dfn>Logic</dfn>: An {{AttributionLogic}}.
-<dfn>Logic Options</dfn>: An {{AttributionLogicOptions}}.
+<dfn>Credit</dfn>: TODO.
```
martinthomson, line 1546: Did we not already have a definition for this?

```suggestion
<dfn>Credit</dfn>: A [=list=] of numbers.
```

martinthomson (commented):
This seems like a good direction.  I'm not sure that we even need to replace the enum if we do the programmatic attribution thing.

Issue #189, opened 2025-05-29
Specify when the Save-Impression HTTP response header is processed


apasel422:
For example, is it processed on potentially every single response, or only a subset? And how do things like redirects work? Do pages have to opt in on the HTML side, like they do in Attribution Reporting (e.g. `<a href="..." attributionsrc>`)? If so, which HTML elements support it (e.g. `<img>`, `<script>` like in ARA)? And are there background requests or only subresource ones?

[ARA supports](https://github.com/WICG/attribution-reporting-api/blob/main/EVENT.md#registering-attribution-sources) source registrations on responses initiated from:
- `<a href="..." attributionsrc>`
- `<area href="..." attributionsrc>`
- `<script src="..." attributionsrc>`
- `<img src="..." attributionsrc>`
- `window.fetch(..., {attributionReporting: { ... }})`
- `window.open(..., ..., 'attributionsrc')`

martinthomson:
Discussion at F2F 2025-06:

Decided not to go with attributionsrc approach, but instead rely more heavily on Permissions Policy and then limit the set of fetch destinations that can invoke the API.  Permissions Policy has a default allow for both of these.  Double check that permissions policy allows for per-site controls in a way that can be used.

NOT "document" (though maybe same origin); uncertain about "frame" and "iframe" (these seem fine, but ARA didn't need them); but we would definitely want "" (the empty string: `<a ping>` and friends), "image", and "script".

On the conversion side, the ability to spend privacy budget is the main risk.  Some people suggested that retagging conversions is hard (very hard).  Tags for ads churn a lot, but conversion tags are once-in-a-lifetime.  It's too hard to re-tag, so the API opens the possibility that random images can exhaust the budget, but that is what Permissions Policy is for (or, in a pinch, CSP).

We could make a different permission policy split for JS vs. HTTP, with different defaults.  Decided not to do that for now.

PR #280, opened 2025-09-16
HTTP API for Measure Conversion

### LINKED ISSUE #31: HTTP API
In deploying ARA, Google discovered that there are a great many intermediaries who are incapable of running JavaScript and many others that prefer not to.  There's a lot of things like tracking pixels, with the HTTP interactions to load those pixels generating measurement signals as a side effect.

Adding HTTP header fields to those requests that trigger the equivalent of the APIs might make the entire system easier for these systems to consume.

There are some serious difficulties involved with this mode of interaction.  Responses will contain instructions to the browser to save impressions or measure conversions.  New requests will be needed in order to deliver the results of those interactions.  That was true for ARA, but the delays meant that new requests were inevitable.  Here, the new interactions would need to be synthesized somehow.  Using a redirect is one possible option.

apasel422:
Fixes #31

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/ppa-api/pull/280.html" title="Last updated on Dec 18, 2025, 1:35 PM UTC (35f3656)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/280/32c72a2...apasel422:35f3656.html" title="Last updated on Dec 18, 2025, 1:35 PM UTC (35f3656)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -2184,7 +2184,7 @@ To <dfn noexport>parse a `Measure-Conversion` header</dfn> given a [=header valu
 1.  Return (|opts|, |reportUrl|).
 
 Issue: Should we allow `http` for |reportUrl|'s [=url/scheme=]?
-Related to https://github.com/w3c/attribution/issues/146.
+Related to <a href="https://github.com/w3c/attribution/issues/146">issue 146</a>.
 
 </div>
 
@@ -2192,10 +2192,15 @@ Related to https://github.com/w3c/attribution/issues/146.
 To <dfn noexport>send a report</dfn> given a [=byte sequence=] |report|,
 a [=URL=] |url|, and an [=environment settings object=] |settings|:
 
-1. [=Assert=]: |url| is a [=potentially trustworthy URL=].
-1. [=Assert=]: |url|'s [=url/scheme=] is "`https`".
-1. Let |headers| be a new [=header list=] containing a [=header=] named `"Content-Type"` whose value is `"application/octet-stream"`.
-1. Let |request| be a new [=request=] with the following properties:
+1.  [=Assert=]: |url| is a [=potentially trustworthy URL=].
+1.  [=Assert=]: |url|'s [=url/scheme=] is "`https`".
+1.  Let |headers| be a new [=header list=] containing a [=header=] named
+    `"Content-Type"` whose value is
+    "<code>[[DAP#name-application-dap-report-medi|application/dap-report]]</code>".
+
+    Note: This will need to be updated if {{AttributionAggregationProtocol}}
+    ever gains a value other than {{AttributionAggregationProtocol/dap-15-histogram}}.
+1.  Let |request| be a new [=request=] with the following properties:
     :   [=request/method=]
     ::  "`POST`"
     :   [=request/URL=]
@@ -2214,27 +2219,9 @@ a [=URL=] |url|, and an [=environment settings object=] |settings|:
     ::  true
     :   [=request/credentials mode=]
     ::  "`omit`"
-1. [=Fetch=] |request|, optionally retrying in the event of an error.
-
-Issue: Should the request use a more specific content-type?
-
-Issue: Should the request's [=request/service-workers mode=] be "`none`" instead
-of the default of "`all`"?
-
-Issue: Should the request's [=request/unsafe-request flag=] be set?
-
-Issue: Should the request's [=request/referrer=] be "`no-referrer`" instead of
-the default of "`client"`?
-
-Issue: Should the request's [=request/destination=] be something other than the
-default of the empty string, e.g. "`report`" or even a new value?
-
-Issue: Should the request's [=request/redirect mode=] be something other than the
-default of "`follow`"?
-
-Issue: Should the request use a `null` client instead, per
-<a href="https://fetch.spec.whatwg.org/#fetch-elsewhere-request">the `fetch`
-specification's guidance</a>?
+    :   [=request/referrer=]
+    ::  |url|
+1.  [=Fetch=] |request|, optionally retrying in the event of an error.
 
 </div>
 
@@ -2286,7 +2273,7 @@ and [=response=] |response|, run these steps:
 Issue: Confirm the desired semantics of [:Save-Impression:] and [:Measure-Conversion:] in the same response.
 
 Issue: Should we allow `http` for |response|'s [=response/URL=]'s [=url/scheme=]?
-Related to https://github.com/w3c/attribution/issues/146.
+Related to <a href="https://github.com/w3c/attribution/issues/146">issue 146</a>.
 
 </div>
 


api.bs
```diff
@@ -1951,37 +1957,184 @@ To <dfn noexport>parse a `Save-Impression` header</dfn> given a [=header value=]
...
+
+\`<dfn http-header><code>Measure-Conversion</code></dfn>\` is a
+[=structured header/dictionary|Dictionary Structured Header=]
+set on a response requesting that the user agent invoke the
+<a method for=Attribution>measureConversion()</a> API.
+
+<div class=example id=ex-measure-conversion-header>
+This is the HTTP equivalent of <a href=#ex-measure-conversion>the JavaScript `measureConversion` example</a>:
+<xmp highlight=http>
+Measure-Conversion: aggregation-service="https://aggregator.example/tee", histogram-size=20, epsilon=1.0, lookback-days=14, impression-sites=("publisher.example" "other.example"), impression-callers=("ad-tech.example"), match-values=(2), credit=(0.25 0.25 0.5), value=3, max-value=7
```
martinthomson, line 2003: The value of `aggregation-service` is not something that can be looked up through this API.  That makes this far less reliable than the JS API, which at least can adapt.

Another thing that needs to be noted clearly is that this doesn't let you have multiple `Measure-Conversion` fields.  If there are multiple conversions, then maybe a redirect chain will allow those to be logged.

martinthomson (commented):
This is the easy part.  Getting the answer is the hard part.

csharrison:
Discussed in the Oct 14 call - let's wait to merge until we have support for sending the report.

apasel422:
> Discussed in the Oct 14 call - let's wait to merge until we have support for sending the report.

I think we need to deal with #189 first.

apasel422:
@martinthomson Please take another look.

impl/src/http.ts
```diff
@@ -98,3 +152,84 @@ export function parseSaveImpressionHeader(
...
+
+  opts.credit = parseInnerList(dict, "credit", (i, value) => {
+    if (typeof value !== "number") {
+      throw new RangeError(`credit[${i}] must be a decimal or an integer`);
+    }
+    return value;
+  });
+
+  opts.value = getPositive32BitUnsignedInteger(dict, "value");
+
```
martinthomson:
```suggestion
```

api.bs
```diff
@@ -2016,6 +2026,201 @@ To <dfn noexport>parse a `Save-Impression` header</dfn> given a [=header value=]
...
+<dl dfn-for=measure-conversion>
+  <dt><dfn noexport><code>aggregation-service</code></dfn></dt>
+  <dd>
+    Value of <a dict-member for=AttributionConversionOptions>aggregationService</a>,
+    a [=structured header/string=]. This key is required.
+  </dd>
+  <dt><dfn noexport><code>epsilon</code></dfn></dt>
+  <dd>
+    Value of <a dict-member for=AttributionConversionOptions>epsilon</a>,
+    a positive [=structured header/decimal=]. This key is optional.
```
```suggestion
    a positive [=structured header/decimal=] or [=structured header/integer=]. This key is optional.
```

api.bs
```diff
@@ -2016,6 +2026,201 @@ To <dfn noexport>parse a `Save-Impression` header</dfn> given a [=header value=]
 
 </div>
 
+## Measuring Conversions ## {#http-api-conversions}
+
+\`<dfn http-header><code>Measure-Conversion</code></dfn>\` is a
+[=structured header/dictionary|Dictionary Structured Header=]
+set on a response requesting that the user agent invoke the
+<a method for=Attribution>measureConversion()</a> API.
```
martinthomson: In line with RFC 9651 we should state:

* unknown dictionary keys are ignored
* unknown parameters on dictionary items are ignored (effectively, all parameters are ignored)

api.bs
```diff
@@ -2016,6 +2026,201 @@ To <dfn noexport>parse a `Save-Impression` header</dfn> given a [=header value=]
...
+    ::  |settings|
+    :   [=request/mode=]
+    ::  "`cors`"
+    :   [=request/cache mode=]
+    ::  "`no-store`"
+    :   [=request/keepalive=]
+    ::  true
+1. [=Fetch=] |request|, optionally retrying in the event of an error.
+
+Issue: Should the request use a more specific content-type?
```
martinthomson: If we're doing DAP only for now, we could hard-code this.  Or we could put the content type into the value we supply and interrogate that.  We could add to [`AttributionConversionResult`](https://w3c.github.io/attribution/#dictdef-attributionconversionresult).

apasel422: The caller of the API is at least theoretically aware of what protocol is being used, even if we support something other than DAP, as the caller chooses the aggregation service, so I'm somewhat reluctant to redundantly include it as part of the request body, which would mean defining some wrapper format around just the encrypted bytes of the report.

In other words, I think whatever we choose should just go in the `Content-Type` header. Does DAP have an existing media type? If not, we could consider registering one but go with `application/octet-stream` for now.

api.bs
```diff
@@ -2016,6 +2026,220 @@ To <dfn noexport>parse a `Save-Impression` header</dfn> given a [=header value=]
...
+    :   [=request/keepalive=]
+    ::  true
+    :   [=request/credentials mode=]
+    ::  "`omit`"
+1. [=Fetch=] |request|, optionally retrying in the event of an error.
+
+Issue: Should the request use a more specific content-type?
+
+Issue: Should the request's [=request/service-workers mode=] be "`none`" instead
+of the default of "`all`"?
```
martinthomson: I would tentatively say no.  Whatever `keepalive` does is fine here.

api.bs
```diff
@@ -2016,6 +2026,220 @@ To <dfn noexport>parse a `Save-Impression` header</dfn> given a [=header value=]
...
+
+Issue: Should the request's [=request/unsafe-request flag=] be set?
```
martinthomson: This is largely moot.  We're using POST with something other than application/x-www-form-urlencoded, so we'll be doing a preflight for any report.  We can't really avoid doing that, and nor should we try.  Enterprises might not like us pushing these reports at random servers, even though the opportunity for attack seems very much remote (no site can really control the content of reports).

api.bs
```diff
@@ -2016,6 +2026,220 @@ To <dfn noexport>parse a `Save-Impression` header</dfn> given a [=header value=]
...
+
+Issue: Should the request's [=request/referrer=] be "`no-referrer`" instead of
+the default of "`client"`?
+
+Issue: Should the request's [=request/destination=] be something other than the
+default of the empty string, e.g. "`report`" or even a new value?
```
martinthomson: Empty seems right.  We'll be keeping good company.  We could attempt to define a destination, but I believe that that is possible to do later, if we think that the destination might help us do something special, like have specific CSP rules.  Right now, I think we're OK relying on permissions policy and the standard connect-src rules from CSP, but we might decide to add a CSP-level control in future for this.

api.bs
```diff
@@ -2016,6 +2026,220 @@ To <dfn noexport>parse a `Save-Impression` header</dfn> given a [=header value=]
...
+
+Issue: Should the request's [=request/redirect mode=] be something other than the
+default of "`follow`"?
```
martinthomson: The default works fine.

api.bs
apasel422: Updated to use `application/dap-report` for now.

PR #288, opened 2025-09-25
Normatively define earliest epoch index in terms of maximum lookback

apasel422:
IMO the benefit of the previous non-normative guidance around how the earliest epoch index might be chosen was unclear, and made it hard to understand what was going on in the simulator with respect to expiry vs lookback. In particular, there was no indication of why implementations might want to vary this value by site, and the non-normative guidance didn't even suggest doing so.

Instead, we normatively define the earliest epoch index in terms of the maximum lookback to permit the full range of impressions to be queried using a maximal `lookbackDays` value.

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/ppa-api/pull/288.html" title="Last updated on Sep 25, 2025, 2:53 PM UTC (7727f1b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/288/26084cd...apasel422:7727f1b.html" title="Last updated on Sep 25, 2025, 2:53 PM UTC (7727f1b)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -791,7 +791,7 @@ The arguments to <a method for=Attribution>measureConversion()</a> are as follow
   <dt><dfn>histogramSize</dfn></dt>
   <dd>The number of histogram buckets to use in the [=conversion report=].</dd>
   <dt><dfn>lookbackDays</dfn></dt>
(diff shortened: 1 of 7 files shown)


csharrison:
This seems fine to me. I don't see any real reason why these should use different constants.

martinthomson (approved):
Much better, thanks.

PR #290, opened 2025-09-30
Add e2e test for Clear-Site-Data integration

apasel422:


api.bs
```diff
@@ -984,36 +984,45 @@ defined in [[CLEAR-SITE-DATA#header]].
...
 
 1.  Let |site| be the value returned
     by invoking [=registrable domain|obtain a registrable domain=],
-    passing the [=host=] part of the [=tuple origin|origin tuple=].
+    passing |origin|'s [=host=] part.
 
 1.  [=set/iterate|For each=] [=impression=] |impression| of
     the [=impression store=]:
 
+    1.  If |impression|'s [=impression/Intermediary Site=] is `undefined` and
```
martinthomson: Why only do this if the intermediary site is undefined?  Shouldn't this happen no matter what the intermediary site is set to?

martinthomson: Let's ask for more input on this one.  I don't think that we've really talked about what it means to clear site data more broadly, but it seems like the impression site has the power to control access to the API, so maybe it should be able to remove all impressions.

A conversion can ignore the save impression caller identity and only focus on the impression site.  So if that is the mode in use, an impression site might want to be able to ensure that impressions saved by intermediaries don't affect things.

Issue #307, opened 2025-11-06
DAP-report construction algorithm doesn't have access to URL for public-key


apasel422:
https://w3c.github.io/attribution/#construct-a-dap-report has:

> Let pkR be the public key of the corresponding role from the [aggregation service](https://w3c.github.io/attribution/#aggregation-service) [HPKE configuration](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-15#section-4.5.1) obtained for the [aggregation service](https://w3c.github.io/attribution/#aggregation-service) indicated by options.[Aggregation Service](https://w3c.github.io/attribution/#validated-conversion-options-aggregation-service).

But `options.Aggregation Service` is just

```webidl
dictionary AttributionAggregationService {
  required DOMString protocol;
};
```

and therefore lacks a URL that would be associated with a public key for encryption.

We should refactor the spec to avoid passing around `AttributionAggregationService` and instead propagate the aggregation service URL itself.

PR #308, opened 2025-11-06
Parse aggregation service as URL and propagate it to DAP-report creation

### LINKED ISSUE #306: Aggregation-service URL lookup uses exact string matching
Currently, the [`validate AttributionConversionOptions` algorithm](https://w3c.github.io/attribution/#validate-attributionconversionoptions) says:

> 1. If [aggregationServices](https://w3c.github.io/attribution/#dom-attribution-aggregationservices) does not [contain](https://infra.spec.whatwg.org/#map-exists) an [entry](https://infra.spec.whatwg.org/#map-entry) with a [key](https://infra.spec.whatwg.org/#map-key) of options.[aggregationService](https://w3c.github.io/attribution/#dom-attributionconversionoptions-aggregationservice), throw a [ReferenceError](https://webidl.spec.whatwg.org/#exceptiondef-referenceerror).
> 2. Let aggregationService be the result of [getting the value](https://infra.spec.whatwg.org/#map-get) from [AttributionAggregationServices](https://w3c.github.io/attribution/#attributionaggregationservices), given options.[aggregationService](https://w3c.github.io/attribution/#dom-attributionconversionoptions-aggregationservice).

However, the IDL defines `aggregationServices` as a `maplike<USVString, AttributionAggregationService>`, which means that keys are compared using string equality, not [URL equivalence](https://url.spec.whatwg.org/#url-equivalence). This means, for example, that `https://EXAMPLE.COM` will not match `https://example.com`.

This seems unlikely to be an issue in practice, but we may want to do something more robust against this kind of canonicalization issue, such as parsing the `AttributionConversionOptions.aggregationService` field as a URL and then looking it up in the user agent's list of allowed aggregation services, which would be a specification-level entity represented as a set of URLs, separate from the `maplike` IDL definition.

[This example in the spec](https://w3c.github.io/attribution/#ex-find-service) seems prone to this kind of issue.

A minor benefit of this would be to align with the treatment of the conversion/impression sites/callers fields, [which are already parsed using the host parser](https://w3c.github.io/attribution/#parse-a-site) rather than compared as strings.

apasel422:
Fixes #306, #307

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/ppa-api/pull/308.html" title="Last updated on Nov 6, 2025, 6:32 PM UTC (804bfed)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/308/2c38eb6...apasel422:804bfed.html" title="Last updated on Nov 6, 2025, 6:32 PM UTC (804bfed)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -472,9 +472,20 @@ This provides three key functions:
 
 ## Finding a Supported Aggregation Service ## {#find-aggregation-service}
 
-The <dfn attribute for=Attribution>aggregationServices</dfn> attribute
-contains a set of aggregation services supported by the [=user agent=]. The page
-must select and specify one of these services when calling the
+The user agent maintains a [=map=] of <dfn>aggregation services</dfn>, whose
+keys are [=URLs=] and whose values are {{AttributionAggregationProtocol}}.
+
+<div algorithm>
+The <dfn attribute for=Attribution>aggregationServices</dfn> getter steps are:
+
+1.  Let |map| be a new [=map=].
+1.  [=map/iterate|For each=] |url| → |protocol| of [=aggregation services=]:
+    1. [=map/Set=] |map|[|url|] to «[ {{AttributionAggregationService/protocol}} → |protocol| ]».
+1.  Return |map|.
+
+</div>
+
+The page must select and specify one of these services when calling the
 <a method for=Attribution>measureConversion()</a> method.
 It may also be useful to query the supported services
 before registering an impression,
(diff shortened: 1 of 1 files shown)


csharrison (approved):

siammoite19855 (approved):
https://github.com/w3c/attribution/tree/main

api.bs
```diff
@@ -2188,7 +2198,7 @@ and a [=list=] of [=integers=] |histogram|:
     1.  Let |pkR| be the public key of the corresponding role from
         the [=aggregation service=] [HPKE configuration](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-15#section-4.5.1)
         obtained for the [=aggregation service=]
-        indicated by |options|.[=validated conversion options/Aggregation Service=].
+        indicated by |options|.[=validated conversion options/Aggregation Service URL=].
```
martinthomson: This means that the URL for the aggregation service points at a configuration.  I guess that's fine, but it also means that it might not be where reports are submitted.

martinthomson (approved):

api.bs
apasel422: This just says "indicated by", meaning there could be an additional mapping maintained by the browser from this URL to some other one for the purpose of key lookup.

PR #321, opened 2025-11-19
Parse aggregation service as URL in simulator

apasel422:
Followup to #308.

### DIFF
--- a/impl/e2e-tests/normalized-aggregation-service-url.json
+++ b/impl/e2e-tests/normalized-aggregation-service-url.json
@@ -1,0 +1,14 @@
+{
+  "events": [
+    {
+      "seconds": 1,
+      "site": "advertiser.example",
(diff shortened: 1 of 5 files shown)


impl/e2e-tests/normalized-aggregation-service-url.json
```diff
@@ -0,0 +1,14 @@
+{
+  "events": [
+    {
+      "seconds": 1,
+      "site": "advertiser.example",
+      "event": "measureConversion",
+      "options": {
+        "aggregationService": "https://AGG-SERVICE.example",
```
martinthomson: Hmm, I'm not sure that I like the way this is turning out.  Shouldn't we be doing simple string comparison for these?  Otherwise, we are exposed to URL normalization for interoperability.

PR #329, opened 2025-12-03
Refine error coverage in e2e tests

apasel422:


martinthomson (approved):
I take this to be due to wanting to increase the distance between invalid ":" and valid "a.example" in site names.

PR #331, opened 2025-12-03
Require site inputs to simulator to already be valid

apasel422:
Rather than parsing those inputs into sites, which is not performed by the corresponding specification algorithms.

### DIFF
--- a/impl/src/backend.ts
+++ b/impl/src/backend.ts
@@ -62,6 +62,12 @@ function parseSite(input: string): string {
   return site;
 }
 
+function validateSite(input: string): void {
+  if (parseSite(input) !== input) {
+    throw new TypeError("input must already be a valid site");
+  }
+}
+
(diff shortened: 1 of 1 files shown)


csharrison (approved):

martinthomson:
Is this right?  Not sure whether this corresponds to the API definitions, which allow someone to say "foo.a.example" which is turned into a site of "a.example".

apasel422:
> Is this right? Not sure whether this corresponds to the API definitions, which allow someone to say "foo.a.example" which is turned into a site of "a.example".

This corresponds to the [implicit inputs](https://w3c.github.io/attribution/#implicit-api-inputs) fields, which are already a site. Unfortunately in the JS code everything is string-ly typed.

martinthomson:
Ah, thanks for pointing that out.  JSON is the best of all possible formats!

thalinda-oss (approved):

PR #332, opened 2025-12-04
Add implementation-defined maximums

### LINKED ISSUE #174: Size checks for lists passed to the API
There are a few places where the API accepts a list of values.  We should:

1. Ensure that each has a validation step that checks against an implementation-defined maximum.
2. Set a minimum value for that maximum, so that sites using the API can provide a list without needing to understand which browser is running and the value that browser picked.

For instance, when saving an impression, there is a list of conversion sites.  At our 2025-05 meeting, it was suggested that a minimum maximum of 10 would be better than the 3 in ARA, which can be too small for some sites.

For conversion sites, an alternative is to omit the list entirely if there is any risk that it might exceed a limit. That makes the impression available on any site, but that has some other unwanted effects, like making the impression invisible to Clear-Site-Data.

There is also a list of match values in the `measureConversion()` arguments.

martinthomson:
This includes a bunch of values that I essentially made up. Rationale for that is listed in #174.
We'll have to discuss what appropriate values are
before we can merge this change.

Closes #174.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/pull/332.html" title="Last updated on Jan 15, 2026, 3:45 AM UTC (d25e2cb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/332/bf7d64a...d25e2cb.html" title="Last updated on Jan 15, 2026, 3:45 AM UTC (d25e2cb)">Diff</a>

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,3 @@
+# Attribution API: https://github.com/w3c/attribution
+
+Copyright © 2026 World Wide Web Consortium. All Rights Reserved. This work is distributed under [the W3C® Software and Document License](https://www.w3.org/Consortium/Legal/copyright-software) in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
--- a/impl/e2e-tests/multi-epoch-budgeting.json
+++ b/impl/e2e-tests/multi-epoch-budgeting.json
@@ -1,0 +1,108 @@
+{
+  "events": [
+    {
+      "$comment": "Associated with epoch=-2 for the conversion sites",
+      "seconds": 1,
+      "site": "publisher.example",
+      "event": "saveImpression",
+      "options": {
+        "histogramIndex": 0,
+        "matchValue": 123
+      }
+    },
+    {
+      "$comment": "Associated with epoch=-1 for the conversion sites",
+      "seconds": 604801,
+      "site": "publisher.example",
+      "event": "saveImpression",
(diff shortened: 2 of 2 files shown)


impl/src/backend.ts
```diff
@@ -68,8 +68,17 @@ function validateSite(input: string): void {
...
 }
 
-function parseSites(input: readonly string[]): Set<string> {
+function parseSites(
+  input: readonly string[],
+  label: string,
+  limit: number,
+): Set<string> {
   const parsed = new Set<string>();
+  if (limit !== null && input.length > limit) {
```
apasel422: The `limit !== null` condition is redundant with `limit: number` above.

impl/src/backend.ts
```diff
@@ -80,10 +89,21 @@ export interface Delegate {
   readonly aggregationServices: AttributionAggregationServices;
   readonly includeUnencryptedHistogram?: boolean;
 
+  /// The maximum number of conversion callers per impression.
   readonly maxConversionSitesPerImpression: number;
+  /// The maximum number of conversion sites per impression.
   readonly maxConversionCallersPerImpression: number;
```
apasel422:
```suggestion
  /// The maximum number of conversion sites per impression.
  readonly maxConversionSitesPerImpression: number;
  /// The maximum number of conversion callers per impression.
  readonly maxConversionCallersPerImpression: number;
```

apasel422:
@martinthomson Just following up on this.

apasel422 (approved):

PR #333, opened 2025-12-04
Add coverage for errors at structured-headers level

apasel422:


impl/src/http.test.ts
```diff
@@ -29,6 +29,9 @@ function runTests(cases: readonly TestCase[]): void {
 }
 
 runTests([
+  { name: "invalid-structured-header-syntax", input: "!" },
+  { name: "not-structured-header-dictionary", input: "histogram-index" },
```
martinthomson:
```suggestion
  { name: "not-structured-header-dictionary", input: "histogram-index" },
  { name: "a-different-type", input: "10" },
```

The "histogram-index" is probably the best test here, because it starts with a label, but that's subtle.

Issue #340, opened 2025-12-09
End-to-end tests lack coverage of user-initiated data clears


apasel422:
Both [with](https://w3c.github.io/attribution/#clear-browsing-history) and [without](https://w3c.github.io/attribution/#clear-site-data) clearing browsing history.

PR #341, opened 2025-12-09
Add e2e tests for user-initiated data clears

apasel422:
#340

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/ppa-api/pull/341.html" title="Last updated on Dec 10, 2025, 1:29 PM UTC (5c7dcc6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/341/a0a78f5...apasel422:5c7dcc6.html" title="Last updated on Dec 10, 2025, 1:29 PM UTC (5c7dcc6)">Diff</a>

### DIFF
--- a/impl/e2e-tests/forget-one-site-conversions.json
+++ b/impl/e2e-tests/forget-one-site-conversions.json
@@ -26,6 +26,13 @@
     },
     {
       "seconds": 4,
+      "site": "a.example",
+      "event": "saveImpression",
+      "$comment": "add an impression to be sure that budget is affected below, rather than impressions",
+      "options": { "histogramIndex": 0 }
+    },
+    {
+      "seconds": 5,
       "site": "advertiser-1.example",
       "event": "measureConversion",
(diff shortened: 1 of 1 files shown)


impl/e2e-tests/clear-site-state.json
```diff
@@ -0,0 +1,51 @@
...
+      "seconds": 4,
+      "site": "advertiser-1.example",
+      "event": "measureConversion",
+      "$comment": "re-run query, expecting budget to have been zeroed",
+      "options": {
+        "aggregationService": "https://agg-service.example",
+        "epsilon": 0.1,
+        "histogramSize": 1
+      },
+      "expected": [0]
```
martinthomson: The reason that this fails is that the site lost its budget.

Would it make sense to have another test where an impression is removed instead?  You would have to advance time by an epoch between the time of the clear and the `measureConversion()` call, but I think that this is doable.

PR #342, opened 2025-12-10
Refine e2e configuration validation

apasel422:


impl/e2e.schema.json
```diff
@@ -83,7 +93,8 @@
                       {
                         "type": "array",
                         "items": {
-                          "type": "integer"
+                          "type": "integer",
+                          "minimum": 0
```
martinthomson: Do you really have to specify a minimum of zero?  I thought that was implied.

martinthomson (approved):

impl/e2e.schema.json
martinthomson: For the length of a list, or is this a per-value constraint?  I'm still a JSON Schema neophyte.

PR #345, opened 2025-12-15
Replace hardcoded "randomness" in e2e tests with configurable value

apasel422:
To make those tests less dependent on fixed behavior.

In the future, we might consider allowing these values to be specified in individual events at a higher level in order to avoid mandating the exact way in which randomness is used.

### DIFF
--- a/impl/e2e-tests/CONFIG.json
+++ b/impl/e2e-tests/CONFIG.json
@@ -2,6 +2,8 @@
   "aggregationServices": {
     "https://agg-service.example": "dap-15-histogram"
   },
+  "epochStart": 0.5,
+  "fairlyAllocateCreditFraction": 0.5,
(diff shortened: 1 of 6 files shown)


apasel422:
> This is a good step, but in my experience a single fixed value doesn't give you what you want longer term. A seeded PRNG tends to be the next step. Then you supply a seed in configuration.

Agreed, though I think this is actually still overly constraining, as it ends up requiring the implementation's PRNG calls to be in some specific order, when the specification itself doesn't necessarily require it.

In the future I'd like to do one of two things instead:
1. Allow tests to provide multiple allowed expected values.
2. Allow tests to supply the value *derived* from the PRNG instead of the random value; this is effectively already the case for the epoch start, as the number corresponds to the location within an epoch, but is not really the case for the fair allocation part.

Issue #43, opened 2024-10-03
Anti-replay requirements


andyleiserson:
This is a TODO section in the spec and was also the subject of some discussion at TPAC.

csharrison:
A few things I recall discussing at TPAC:
1. Grouping together the state associated with multiple reports via origin / timestamp for efficiency (we do this in ARA)
2. Adding some associated data in the AEAD that allows the API caller to further partition the groups in a finer-grained way
3. Whether to support requerying the same report via the stateful anti-replay protection (vs. just emitting multiple reports).

IMO I think (2) and (3) are probably puntable from level 1 since it seems fairly easy to extend in the future, but we should specify (1) at the minimum.

martinthomson:
This will be addressed by DAP.  That's good for us, because anti-replay is a giant PITA.

PR #354, opened 2026-01-05
anti-replay requirements for level 1

bmcase:
anti-replay requirements -- just the basic ones from https://github.com/w3c/attribution/issues/43 not including partitioning and requerying.

I have a separate branch with addition of associated data in the AEAD to partition reports but requires more changes and probably can wait till level 2 as @csharrison mentioned on the issue.

***
<a href="https://pr-preview.s3.amazonaws.com/bmcase/api/pull/354.html" title="Last updated on Jan 5, 2026, 4:49 PM UTC (3ade7fe)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/354/e4f71d3...bmcase:3ade7fe.html" title="Last updated on Jan 5, 2026, 4:49 PM UTC (3ade7fe)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -2498,7 +2498,7 @@ and a [=list=] of [=integers=] |histogram|:
         mapped to the [=isomorphic encode|encoded=] value of |topLevelSite|[1].
 
 1.  Let |reportMetadata| be encoded DAP [`ReportMetadata`](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-15#section-4.5.2)
-    generated from |reportID|, |time|, and |extensions|.
+    generated from |reportID|, |topLevelSite|, |time|, and |extensions|.
 
 1.  Let |encryptedInputShares| be an [=list/is empty|empty=] [=list=].
 
--- a/api.bs
+++ b/api.bs
@@ -2555,6 +2555,20 @@ that was expended by the site that requested the report.
 An [=aggregation service=] MUST guarantee
 that it does not accept the same report more than once.
 
+To support anti-replay verification,
+each [=conversion report=] includes associated data
+that is authenticated as part of the AEAD encryption.
+This associated data includes:
+
+*   The [=origin=] of the caller invoking the {{Attribution/measureConversion()}} API.
+*   A timestamp, as determined by the device at the time of the API call.
+
+The [=aggregation service=] uses this associated data
+to shard its anti-replay mechanism.
+This sharding enables efficient verification
+that each report is processed only once,
+while allowing the service to scale.
+
 
 # Differential Privacy # {#dp}
 
--- a/api.bs
+++ b/api.bs
@@ -2560,7 +2560,7 @@ each [=conversion report=] includes associated data
 that is authenticated as part of the AEAD encryption.
 This associated data includes:
 
-*   The [=origin=] of the caller invoking the {{Attribution/measureConversion()}} API.
+*   The [=site=] of the caller invoking the {{Attribution/measureConversion()}} API.
 *   A timestamp, as determined by the device at the time of the API call.
 
 The [=aggregation service=] uses this associated data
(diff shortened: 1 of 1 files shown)


frankquinonez681:
I want peace and understanding of whats going on

frankquinonez681 (commented):
Thank you, not much should change

api.bs
```diff
@@ -2498,7 +2498,7 @@ and a [=list=] of [=integers=] |histogram|:
         mapped to the [=isomorphic encode|encoded=] value of |topLevelSite|[1].
 
 1.  Let |reportMetadata| be encoded DAP [`ReportMetadata`](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-15#section-4.5.2)
-    generated from |reportID|, |time|, and |extensions|.
+    generated from |reportID|, |topLevelSite|, |time|, and |extensions|.
```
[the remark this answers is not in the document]
csharrison: But "site of the caller" vs. `topLevelSite` is a different concept. Which one should it be? I believe this one should be `site` and not `topLevelSite`.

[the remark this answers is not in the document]
bmcase: Agree, it should be the site of the caller.  Updated to define a |conversionCaller| like is done in the Common Impression Matching Logic section https://w3c.github.io/attribution/#logic-matching

api.bs
```diff
@@ -2555,6 +2560,22 @@ that was expended by the site that requested the report.
...
 that it does not accept the same report more than once.
 
+To support anti-replay verification,
+each [=conversion report=] includes associated data
+that is authenticated as part of the AEAD encryption.
+This associated data includes:
+
+*   The [=site=] that invoked the {{Attribution/measureConversion()}} API.
+    This is the [=intermediary site=] if the API was called from a cross-site frame,
+    or the [=conversion site=] otherwise.
```
martinthomson: This relates to the changes you added to the algorithms, but I don't think that this is correct.

We charge budget to the top-level site.  Therefore the aggregation service needs to attribute individual reports to the same entity.  Otherwise, we create the possibility that reports from multiple sites (and therefore multiple budgets) are aggregated together and noise is not correctly applied.

Concretely, if I have reports from site A and site B, then there are two budgets. I would expect two lots of noise as well.  This approach would allow an intermediary to combine reports from sites into a single aggregation that includes just one lot of noise.

martinthomson (changes requested):
I might be missing something, but I don't think that this is right.

api.bs
csharrison: I think we should discuss this next week. Here is my opinion:

1. The conversion caller should be in the associated data and we should fail decryption in the helpers if the callers don't match. This prevents report-theft from being a problem.
2. I don't think the top level site needs to be in the associated data. I don't think a single lot of noise is problematic for the guarantees (esp. if safety limits are in place), but @bmcase can correct me if I'm wrong. In fact, I think we wanted to support cross-advertiser queries right?

bmcase: discussed with @csharrison; we think we should include the following

1. the caller of the API (either an intermediary (like measurement provider for the advertiser) or the advertiser themself) and for this use the full origin.   This can be used to make sure only this origin can query the report to the Helpers (and prevent DOS on reports being queried by some other party). 

1. the eTLD+1 of the conversion site.  This corresponds to the grain of the DP budgeting in Level 1 and prevents aggregation of reports from different advertisers which spent different budgets.  (This can be handled different in Level 2 when we do support cross-advertiser aggregation and there ensure we bind to whichever budget is being deducted from).

csharrison:
Editor discussion 2/5: we want to ensure the caller and top level site are in the report. Caller is needed just to prevent report stealing (somewhat orthogonal to anti-replay).

We aligned it is probably OK to bind this to the task.

bmcase:
Closing this as we mentioned above these changes will be covered more completely in the DAP integration work.

Issue #351, opened 2025-12-18
End-to-end tests lack coverage for budget exhaustion


apasel422:
Corresponding to Step 10.5 of https://w3c.github.io/attribution/#do-attribution-and-fill-a-histogram:

> If `budgetOk` is false, set `histogram` to the result of invoking create an all-zero histogram with `options`’ histogram size.

Issue #336, opened 2025-12-04
Overly conservative check and deduct for budget


bmcase:
The current text in [=deduct privacy budget=] function is a bit overly conservative when checking and deducting budget

> 1.  If |deductionFp| is negative or greater than [=maximum epsilon=], [=map/set|set=] the value of |key| in the [=privacy budget store=] to 0
    and return false.

Also

> 1.  If |deduction| is greater than |currentValue|, [=map/set|set=] the value of |key| in the [=privacy budget store=] to 0
    and return false.

If someone asks for a really big epsilon, you don’t have to both zero out their budget and give them a null report. You can just give them a null report and hope they figure out how this API works to come back and spend their budget properly… 

I think I'll end up addressing this in https://github.com/w3c/attribution/pull/309 as need to also make it such that privacy budget deduction occurs if and only if all safety limit deductions can be made, otherwise just drop the epoch's impressions from consideration in this measureConversion request.

martinthomson:
Are you fixing this in #309, or do you want to take a stab at fixing this @bmcase ?

PR #356, opened 2026-01-06
Add e2e test for single-epoch budgeting

apasel422:
Part of #351.

### DIFF
--- a/impl/e2e-tests/single-epoch-budgeting.json
+++ b/impl/e2e-tests/single-epoch-budgeting.json
@@ -1,0 +1,119 @@
+{
+  "events": [
+    {
+      "seconds": 1,
+      "site": "publisher.example",
+      "event": "saveImpression",
+      "options": { "histogramIndex": 0 }
+    },
+    {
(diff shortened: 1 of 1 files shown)


impl/e2e-tests/single-epoch-budgeting.json
```diff
@@ -0,0 +1,119 @@
...
+      "options": {
+        "aggregationService": "https://agg-service.example",
+        "epsilon": 1,
+        "histogramSize": 3,
+        "lookbackDays": 1,
+        "value": 4,
+        "maxValue": 8,
+        "credit": [3, 1]
+      },
+      "$comment": "The preceding case should have subtracted the remaining 1/4 even though it was insufficient for the 1/2 required",
```
csharrison: https://github.com/w3c/attribution/issues/336 may change this (non-blocking)

apasel422: Oh, nice. Glad we'll have evidence for the different behavior before and after.

PR #370, opened 2026-02-10
Add SLSA generic generator workflow

ravikantcool2024:
This workflow generates SLSA provenance files for projects, satisfying level 3 requirements. It includes steps for building artifacts and generating provenance subjects.


bmcase:
closing as appears to be spam.

PR #309, opened 2025-11-12
[RFC] Safety limits

bmcase:
Creating a PR to add safety limits to the Attribution spec.  This is based primarily on the BigBird algorithm from Section 4 of this paper https://arxiv.org/pdf/2506.05290.  Algorithm 2 is the main algorithm that encompass both budget deduction and safety limit deduction.  

Intended to address this open issue https://github.com/w3c/attribution/issues/237

***
<a href="https://pr-preview.s3.amazonaws.com/bmcase/api/pull/309.html" title="Last updated on Apr 20, 2026, 7:51 PM UTC (646504c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/309/c0cf215...bmcase:646504c.html" title="Last updated on Apr 20, 2026, 7:51 PM UTC (646504c)">Diff</a>

PR #376, opened 2026-02-11
Add safety limits to simulator

apasel422:
Corresponding to #309 and #403.

### DIFF
--- a/impl/e2e-tests/CONFIG.json
+++ b/impl/e2e-tests/CONFIG.json
@@ -8,6 +8,8 @@
   },
   "epochStart": 0.5,
   "fairlyAllocateCreditFraction": 0.5,
+  "globalPrivacyBudgetPerEpoch": 8000000,
+  "impressionSiteQuotaPerEpoch": 4000000,
   "maxConversionSitesPerImpression": 3,
   "maxConversionCallersPerImpression": 3,
   "maxImpressionSitesForConversion": 3,
@@ -16,6 +18,6 @@
   "maxMatchValues": 10,
   "maxLookbackDays": 30,
   "maxHistogramSize": 5,
-  "privacyBudgetMicroEpsilons": 1000000,
+  "perSitePrivacyBudget": 1000000,
   "privacyBudgetEpochDays": 7
 }
--- a/impl/e2e-tests/single-epoch-budgeting.json
+++ b/impl/e2e-tests/single-epoch-budgeting.json
@@ -73,8 +73,8 @@
         "maxValue": 8,
         "credit": [3, 1]
       },
-      "$comment": "The preceding case should have subtracted the remaining 1/4 even though it was insufficient for the 1/2 required",
-      "expected": [0, 0, 0]
+      "$comment": "Budget was not consumed by the preceding case (insufficient budget causes no deduction), so 1/4 remains which is exactly the 1/4 required here",
+      "expected": [1, 3, 0]
     },
     {
       "seconds": 7,
--- a/impl/src/backend.ts
+++ b/impl/src/backend.ts
@@ -47,6 +47,8 @@ interface ValidatedConversionOptions {
   maxValue: number;
 }
 
+const UNIX_EPOCH = new Temporal.Instant(0n);
+
 export function days(days: number): Temporal.Duration {
   // We use `hours: X` here instead of `days` because days are considered to be
   // "calendar" units, making them incapable of being used in calculations
@@ -103,8 +105,10 @@ export interface Delegate {
   readonly maxLookbackDays: number;
   /// The maximum size of histograms.
   readonly maxHistogramSize: number;
-  readonly privacyBudgetMicroEpsilons: number;
+  readonly perSitePrivacyBudget: number;
   readonly privacyBudgetEpoch: Temporal.Duration;
+  readonly impressionSiteQuotaPerEpoch: number;
+  readonly globalPrivacyBudgetPerEpoch: number;
 
   now(): Temporal.Instant;
   fairlyAllocateCreditFraction(): number;
@@ -133,22 +137,43 @@ function isDouble(v: number): boolean {
   return Number.isFinite(v);
 }
 
+function getEntry<E extends Readonly<PrivacyBudgetKey>>(
+  store: E[],
+  key: Readonly<PrivacyBudgetKey>,
+): E | undefined {
+  return store.find((e) => e.epoch === key.epoch && e.site === key.site);
+}
+
+function getOrInsertEntry(
+  store: PrivacyBudgetStoreEntry[],
+  key: Readonly<PrivacyBudgetKey>,
+  defaultValue: number,
+): PrivacyBudgetStoreEntry {
+  let entry = getEntry(store, key);
+  if (entry === undefined) {
+    entry = { value: defaultValue, ...key };
+    store.push(entry);
+  }
+  return entry;
+}
+
 export class Backend {
   enabled: boolean = true;
 
   readonly #delegate: Delegate;
   #impressions: Readonly<Impression>[] = [];
-  readonly #epochStartStore: Map<string, Temporal.Instant> = new Map();
+  #epochStartTime: Temporal.Instant | null = null;
+  #globalPrivacyBudgetStore: Map<number, number> = new Map();
   #privacyBudgetStore: PrivacyBudgetStoreEntry[] = [];
-
+  #impressionSiteQuotaStore: PrivacyBudgetStoreEntry[] = [];
   #lastBrowsingHistoryClear: Temporal.Instant | null = null;
 
   constructor(delegate: Delegate) {
     this.#delegate = delegate;
   }
 
-  get epochStarts(): ReadonlyMap<string, Temporal.Instant> {
-    return this.#epochStartStore;
+  get epochStartTime(): Temporal.Instant | null {
+    return this.#epochStartTime;
   }
 
   get privacyBudgetEntries(): ReadonlyArray<Readonly<PrivacyBudgetStoreEntry>> {
@@ -372,10 +397,7 @@ export class Backend {
     const matching = new Set<Impression>();
 
     for (const impression of this.#impressions) {
-      const impressionEpoch = this.#getCurrentEpoch(
-        topLevelSite,
-        impression.timestamp,
-      );
+      const impressionEpoch = this.#getCurrentEpoch(impression.timestamp);
       if (impressionEpoch !== epoch) {
         continue;
       }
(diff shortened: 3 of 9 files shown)


impl/src/backend.ts
```diff
@@ -431,45 +431,62 @@ export class Backend {
...
+    const currentEpoch = this.#getCurrentEpoch(now);
+    const startEpoch = this.#getStartEpoch(now);
+    const earliestEpoch = this.#getCurrentEpoch(now.subtract(options.lookback));
+    const isSingleEpoch = currentEpoch === earliestEpoch;
+    let l1Norm = 0;
 
-    if (singleEpoch) {
+    if (isSingleEpoch) {
       matchedImpressions = this.#commonMatchingLogic(
         topLevelSite,
```
bmcase, line 467: if we want to keep the order of parameters to the function the same as the text spec, it would have `options` first.

impl/src/backend.ts
bmcase, line 300: we should also check value > 0.   We don't want to accept negative values for it.

impl/src/backend.ts
```diff
@@ -431,45 +431,62 @@ export class Backend {
...
+      if (matchedImpressions.size === 0) {
+        return allZeroHistogram(options.histogramSize);
+      }
+      const histogram = this.#fillHistogramWithLastNTouchAttribution(
+        matchedImpressions,
+        options.histogramSize,
+        options.value,
+        options.credit,
+      );
+      l1Norm = histogram.reduce((a, b) => a + b, 0);
```
bmcase, line 482: it should be the case that there are no negative values assigned in the histogram, but maybe we can add a comment here about that so it is easy to follow why we are not also computing absolute value when computing the L1 norm.

impl/src/backend.ts
```diff
@@ -431,45 +431,62 @@ export class Backend {
...
-          intermediarySite,
-          epoch,
-          now,
-          options,
+    }
+
+    const deductedImpressionQuotas: PrivacyBudgetKey[] = [];
+    const deductedGlobalBudgets = new Set<number>();
+
+    for (let epoch = startEpoch; epoch <= currentEpoch; ++epoch) {
```
bmcase, line 490: currently in the spec we loop backwards from current.  I don't think it will make a difference with how we are doing things right now as every epoch is independent but if there were ever any priority I think we would prefer more recent epochs be processed first.

impl/src/backend.ts
```diff
@@ -431,45 +431,62 @@ export class Backend {
...
+      const impressions = this.#commonMatchingLogic(
+        topLevelSite,
```
bmcase, line 492: similar to above comment -- `options` is the first input parameter as listed in the spec if we want to keep that consistent.

impl/src/backend.ts
```diff
@@ -479,68 +496,128 @@ export class Backend {
...
+        (e) =>
+          e.epoch === impressionQuotaKey.epoch &&
+          e.site === impressionQuotaKey.site,
+      );
+      if (entryI === undefined) {
+        entryI = {
+          value: this.#delegate.impressionSiteQuotaPerEpoch,
+          ...key,
+        };
+        this.#impressionSiteQuotaStore.push(entry);
```
bmcase:
```suggestion
        this.#impressionSiteQuotaStore.push(entryI);
```

impl/src/backend.ts
```diff
@@ -479,68 +496,128 @@ export class Backend {
...
+      }
+      if (
+        deductedImpressionQuotas.find(
+          (e) =>
+            e.epoch === impressionQuotaKey.epoch &&
+            e.site === impressionQuotaKey.site,
+        ) === undefined
+      ) {
+        deductedImpressionQuotas.push(impressionQuotaKey);
+        entryI.value -= valueDeduction;
```
bmcase, line 588: Just to confirm, this is going to work as a pointer to update in the `impressionSiteQuotaPerEpoch` store, and that is why we don't use set something like this? 
      `this.#impressionSiteQuotaPerEpoch.set(epoch, entryI.value - valueDeduction);

bmcase (commented):
checked over the main privacy budget related functions and compared with the safety limits PR.   Looks good; left a couple comments.

impl/e2e-tests/CONFIG.json
```diff
@@ -8,6 +8,8 @@
   },
   "epochStart": 0.5,
   "fairlyAllocateCreditFraction": 0.5,
+  "globalPrivacyBudgetPerEpoch": 1000000,
+  "impressionSiteQuotaPerEpoch": 1000000,
```
bmcase: for testing, we should probably make this some multiple of the perSitePrivacyBudget, otherwise tests with multiple advertisers will likely fail.  I see single-epoch-budgeting.json is failing and has two advertisers; I'm not certain this is the reason for that particular test starting to fail but seems it could be from these parameters.

impl/src/backend.ts
apasel422: That is already checked by the "type-level" assertions at the top of the method.

impl/src/backend.ts
apasel422: Unrelated to this PR; will address separately.

impl/src/backend.ts
apasel422: Oh, I missed that in the spec PR. [Left a comment asking about that.](https://github.com/w3c/attribution/pull/309/changes#r3079432795)

impl/src/backend.ts
apasel422: Not related to this PR.

impl/src/backend.ts
apasel422: Good catch.

impl/src/backend.ts
apasel422: Yes.

impl/e2e-tests/CONFIG.json
apasel422: Yes, if you have test suggestions, please feel free to provide them.

impl/src/backend.ts
apasel422: I don't see anything about absolute value in the specification.

impl/e2e-tests/CONFIG.json
apasel422: Quadrupling those two values isn't affecting whether that test passes or fails.

bmcase: @apasel422  you're right it's not the values themselves.  I've been running the simulator and tests and I think I know why that test single-epoch-budgeting.json is failing. 

1. In the single epoch case there needs to be another line to clear matchedImpressions. `matchedImpressions = new Set<Impression>();` at backend.ts 490.   That will fix the test failing at seconds 5. 
2. the test will then start breaking at seconds = 6, that is because the test is assuming overly conservative budget deduction and expecting [0, 0, 0] but should now be expecting [1, 3, 0]

impl/src/backend.ts
```diff
@@ -435,45 +437,62 @@ export class Backend {
...
+        return allZeroHistogram(options.histogramSize);
+      }
+      const histogram = this.#fillHistogramWithLastNTouchAttribution(
+        matchedImpressions,
+        options.histogramSize,
+        options.value,
+        options.credit,
+      );
+      l1Norm = histogram.reduce((a, b) => a + b, 0);
+      assert(l1Norm <= options.value);
```
bmcase, line 483:
```suggestion
      assert(l1Norm <= options.value);
      matchedImpressions = new Set<Impression>();

```

impl/e2e-tests/CONFIG.json
apasel422: Where is the equivalent clear in the spec?

https://github.com/w3c/attribution/blob/904ab26fbb7d56d35c698527f7bf84f3a28f76d8/api.bs#L1886-L1897

impl/src/backend.ts
```diff
@@ -483,68 +502,128 @@ export class Backend {
...
+      const impressionQuotaKey = { site: impressionSite, epoch };
+      let entryI = this.#impressionSiteQuotaStore.find(
+        (e) =>
+          e.epoch === impressionQuotaKey.epoch &&
+          e.site === impressionQuotaKey.site,
+      );
+      if (entryI === undefined) {
+        entryI = {
+          value: this.#delegate.impressionSiteQuotaPerEpoch,
+          ...key,
```
bmcase: @apasel422  I think this `key` should be `impressionQuotaKey`

impl/src/backend.ts
apasel422: If `matchedImpressions` is supposed to be cleared after this, then I think we should just use a different variable in this branch, to make the scoping more obvious. IIUC the spec does not currently have the behavior you are suggesting.

impl/e2e-tests/CONFIG.json
bmcase: You're right, we should have it cleared there in the spec as well. The issue is the first computation with matched impressions is really just computing the L1 norm. After that we want to clear it otherwise there may be data left in it when we go on to the forward loop over epochs and if the budget checks don't pass we will not add anything to it but it could still have entries it in if we don't clear it here.

apasel422: If you put together a PR updating the spec, I'll approve it.

impl/src/backend.ts
bmcase: after making this change many other tests began failing as the impression quota was not otherwise getting applied.   To make those other tests pass we can set the default values for impression quota to 4 and global quota to 8.  

That will keep other tests passing as before; I'm working on a new test that will exhaust the impression quota in certain cases and make sure that does fail when expected.

impl/e2e-tests/CONFIG.json
bmcase: created a change for the spec https://github.com/w3c/attribution/pull/407

impl/src/backend.ts
bmcase: change in spec also: https://github.com/w3c/attribution/pull/407

impl/e2e-tests/CONFIG.json
apasel422: With all your suggestions, single-epoch-budgeting is still failing at seconds=4.

impl/e2e-tests/single-epoch-budgeting.json
```diff
@@ -73,8 +73,8 @@
         "maxValue": 8,
         "credit": [3, 1]
       },
-      "$comment": "The preceding case should have subtracted the remaining 1/4 even though it was insufficient for the 1/2 required",
-      "expected": [0, 0, 0]
+      "$comment": "FIXME: Explain the expectation here",
```
bmcase: "$comment": "Budget was not consumed by the preceding case (insufficient budget causes no deduction), so 1/4 remains which is exactly the 1/4 required here",

bmcase (approved):
this is looking good to me; we should still add some more tests that push against the edges of the updated privacy budgeting logic but we can probably follow up with those in a separate PR.

impl/src/backend.ts
```diff
@@ -483,68 +500,128 @@ export class Backend {
...
+          e.epoch === impressionQuotaKey.epoch &&
+          e.site === impressionQuotaKey.site,
+      );
+      if (entryI === undefined) {
+        entryI = {
+          value: this.#delegate.impressionSiteQuotaPerEpoch,
+          ...impressionQuotaKey,
+        };
+        this.#impressionSiteQuotaStore.push(entryI);
+      }
```
martinthomson: This code repeats with the code to find the per-site budget item.  Is there any value to be gained from factoring this out?

As for the structure of this, this code won't be able to perfectly replicate the get-with-default mode that is part of the spec, but I think that you can factor it out such that it looks closer to that.

```js
const deduct = (store, key, dflt, deduction) => {
  let i = store.find(e => e.epoch === key.epoch && e.site === key.site);
  if (i === undefined) {
    i = { value: dflt, ...key };
    store.push(i);
  }
  i.value -= deduction;
};
```

impl/src/backend.ts
```diff
@@ -483,68 +500,128 @@ export class Backend {
...
+      if (
+        deductedImpressionQuotas.find(
+          (e) =>
+            e.epoch === impressionQuotaKey.epoch &&
+            e.site === impressionQuotaKey.site,
+        ) === undefined
+      ) {
```
martinthomson, line 586: This check seems to be something you could do before you went looking for the entry in the quota store.  That is:

```js
for (const impression of impressions) {
  ...
  if (deductedImpressionQuotas.find(key)) { // obviously, this needs to be expanded, but see below about factoring of this find() invocation
    continue;
  }
  deduct(this.#impressionSiteQuotaStore, impressionQuotaKey, this.#delegate.impressionSiteQuotaPerEpoch, valueDeduction);
}
```

impl/src/backend.ts
```diff
@@ -483,68 +500,128 @@ export class Backend {
...
-    entry.value -= deduction;
+    for (const impression of impressions) {
+      const impressionSite = impression.impressionSite;
+      const impressionQuotaKey = { site: impressionSite, epoch };
+      const currentImpressionQuotaValue =
+        this.#impressionSiteQuotaStore.find(
+          (e) =>
+            e.epoch === impressionQuotaKey.epoch &&
+            e.site === impressionQuotaKey.site,
+        )?.value ?? this.#delegate.impressionSiteQuotaPerEpoch;
```
martinthomson: It seems like this find operation is something that might be worth factoring out.

impl/src/backend.ts
```diff
@@ -692,17 +766,20 @@ export class Backend {
...
       this.#privacyBudgetStore = this.#privacyBudgetStore.filter((e) => {
         return !parsedSites.has(e.site);
       });
-      for (const site of parsedSites) {
-        this.#epochStartStore.delete(site);
-      }
+      this.#impressionSiteQuotaStore = this.#impressionSiteQuotaStore.filter(
+        (e) => {
+          return !parsedSites.has(e.site);
+        },
```
martinthomson:
```suggestion
        (e) => !parsedSites.has(e.site),
```

I take it that this shorter form isn't compatible with prettier.

impl/src/backend.ts
```diff
@@ -483,68 +500,128 @@ export class Backend {
...
+    impressions: ReadonlySet<Impression>,
     epsilon: number,
     value: number,
     maxValue: number,
-    l1Norm: number | null,
+    isSingleEpoch: boolean,
+    l1Norm: number,
+    deductedImpressionQuotas: PrivacyBudgetKey[],
+    deductedGlobalBudgets: Set<number>,
   ): boolean {
```
martinthomson, line 543: Were this not so faithful a reproduction of the algorithm in the spec, I'd be looking for ways to deal with the growing argument list here.  It's possible that we can do some factoring of the spec in tandem with changes here, but that can be deferred.

martinthomson (approved):
Looks like a pretty direct translation to me.  A few comments, but mostly down to your discretion about how you deal with them.

impl/src/backend.ts
apasel422: Done, though I had to exclude the deduction part, as it's only conditionally deducted below.

PR #384, opened 2026-02-18
Track DAP changes

### LINKED ISSUE #35: DAP Integration: sensitivity and privacy budget expenditure
Our design requires that we communicate the sensitivity and budget expenditure that every contribution makes.  The aggregation service needs to take that information and use that in applying DP noise.  There are a few ways to slice this, but my suggestion would be:

1. Each task is configured with a preset sensitivity ($\Delta$) and budget ($\epsilon$).
2. Each submission is bound to these values.  We can use the additional authenticated data (AAD) component of the AEAD we use for encryption, so that the values -- which are public -- don't need to be decrypted and therefore don't need to be resubmitted with every submission.  That will save a lot of overhead for batched submissions.
3. A submission is invalid unless the sensitivity is less than or equal to the configured sensitivity of the query.  Ideally, the values would be the same, which might simplify things.  However, queries will likely need some way to vary metadata for individual submissions, so this might not result in a genuine simplification of the protocol.  We should probably start with the simple approach, though there is no technical or privacy reason that a smaller sensitivity wouldn't work.  Using a smaller sensitivity is going to result in greater noise relative to that contribution, but the smaller sensitivity will reduce the size of the submission, which could be well worth the modest increase in complexity.
4. A submission is invalid unless the privacy budget expended is greater than or equal to the privacy budget of the query.  Again, it would be ideal of the values were the same, with similar effect to that of having a lower sensitivity.
5. The aggregation service would apply noise based on the configured sensitivity and budget.

### LINKED ISSUE #361: Include `maxValue` in report authenticated data
Right now the [encrypted report](https://w3c.github.io/attribution/#construct-a-dap-report 
) only contains epsilon.  

It should also include `maxValue` as we [assume the noise](https://w3c.github.io/attribution/#s-privacy-budget-store ) is added by the Helpers knowing `2 * maxValue / epsilon`

TODO: make sure this can also work with @martinthomson  DAP extension for binding the additional DP data

### LINKED ISSUE #365: Confirm fixed time resolution in DAP integration
https://github.com/w3c/attribution/blob/bf7d64a6426fa6f6603a0afeeb0fe11cd94d9f98/api.bs#L2477-L2479

### LINKED ISSUE #375: Update for DAP changes
After discussions here and with some PPM working group contributors (i.e., most of them), it seems like we have a solid plan for how to integrate with DAP.

The overall thing is blocked on changes to DAP and the taskprov spec.  That work is in progress.  I've put a spec that documents [extensions needed by attribution API for DAP](https://martinthomson.github.io/dap-attribution/draft-thomson-ppm-dap-attribution.html) and will be asking for that to be adopted in the PPM group.  However, all of that is blocked on DAP being revised with the agreed changes.

Current state as of 2026-02-11 is that this is within a couple of weeks of happening:
BACKLOG Tim, J.C.: Move TaskConfig over to dap
BLOCKED MT: Make task lifetime a TaskExtension
[IN PROGRESS](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/769) MT: Specify collection extension, along the same lines as upload extension
[IN PROGRESS](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/766) MT: Add verification key ID to aggregation job

After those start to stabilize, the DAP report creation process in the spec will need to be revised.  I'm happy to do that.

martinthomson:
This is a fairly big overhaul of how the DAP integration works.

Most of the rationale has been discussed already in various issues.

Draft because dap-18 isn't out yet.

Closes #375.
Closes #365.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/pull/384.html" title="Last updated on May 12, 2026, 2:28 PM UTC (8263901)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/384/4683655...8263901.html" title="Last updated on May 12, 2026, 2:28 PM UTC (8263901)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -2568,8 +2568,9 @@ and a [=list=] of [=integers=] |histogram|:
     with |ctx|, |histogram|, |reportID| (as the VDAF "nonce" parameter), and |rand|.
 
 1.  Let |time| be the integer resulting from dividing
-    |now| as a [=duration=] since the [=Unix epoch=]
-    by a [=duration=] of 5 seconds.
+    the [=duration from=] the [=Unix epoch=] to |now|
+    by a [=duration=] of 5 seconds,
+    rounding towards negative Infinity.
 
 1.  Let |extensions| be a [=map=] of [=16-bit unsigned integers=] to [=byte sequences=],
     comprised of:
--- a/api.bs
+++ b/api.bs
@@ -508,7 +508,7 @@ and impressions are not scoped to a single aggregation service.
 </div>
 
 <xmp class=idl>
-enum AttributionAggregationProtocol { "dap-15-histogram" };
+enum AttributionAggregationProtocol { "dap-18-histogram" };
 
 dictionary AttributionAggregationService {
   required AttributionAggregationProtocol protocol;
@@ -548,7 +548,7 @@ The <dfn enum>AttributionAggregationProtocol</dfn> describes the submission prot
 used by different [=aggregation services=]. This document defines two protocols:
 
 <dl dfn-for=AttributionAggregationProtocol dfn-type=enum-value>
-  <dt><dfn>dap-15-histogram</dfn></dt>
+  <dt><dfn>dap-18-histogram</dfn></dt>
   <dd>A DAP-based protocol [[DAP]] that uses [=MPC=]; see [[#s-mpc]].</dd>
 </dl>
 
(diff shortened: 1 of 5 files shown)


api.bs
```diff
@@ -2481,110 +2477,151 @@ To <dfn>construct a DAP report</dfn>,
...
+        taken from the [=implementation-defined=] definition
+        of the selected [=aggregation service=]
+        in |options|.[=validated conversion options/Aggregation Service=],
+        produced using the same process as the `leader_aggregator_endpoint` value.
+
+    1.  A `time_precision` value of 5.
+
+    1.  A `min_batch_size` value of 20.
+
+    1.  A `batch_mode` value of  TBD (see [[DAP-ATTRIBUTION#batch-mode]]).
```
```suggestion
    1.  A `batch_mode` value of TBD (see [[DAP-ATTRIBUTION#batch-mode]]).
```

apasel422 (changes requested):
We should also update the simulator's [uses of the old enum](https://github.com/search?q=repo%3Aw3c%2Fattribution%20path%3A%2F%5Eimpl%5C%2F%2F%20dap-15&type=code).

bmcase (approved):
looks good to me using the extensions from DAP-Attribution

PR #391, opened 2026-03-04
Remove unnecessary use of async/await in clear-data tests

apasel422:
Ultimately, `Backend.saveImpression` is synchronous, so none of the async support for this is necessary. Found by an updated `npm lint`.


csharrison (approved):
Nice, thanks.

PR #394, opened 2026-03-09
Replace unmaintained psl dependency with tldts

apasel422:
https://github.com/remusao/tldts?tab=readme-ov-file#migrating-from-other-libraries

### DIFF
--- a/impl/src/backend.ts
+++ b/impl/src/backend.ts
@@ -13,7 +13,7 @@ import assert from "assert";
 
 import { Temporal } from "temporal-polyfill";
 
-import * as psl from "psl";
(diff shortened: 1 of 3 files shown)


impl/src/backend.ts
```diff
@@ -13,7 +13,7 @@ import assert from "assert";
 
 import { Temporal } from "temporal-polyfill";
 
-import * as psl from "psl";
+import { getDomain } from "tldts";
```
martinthomson, line 16: Hmm, that's a bad name for this function.

martinthomson (approved):

PR #400, opened 2026-04-08
Replace production assert dependency with simple implementation

apasel422:
It was causing "process not found" errors to be thrown, breaking the simulator at https://w3c.github.io/attribution/simulator.html.


csharrison (approved):
Amazing

PR #403, opened 2026-04-20
Clearing state for safety limits

### LINKED ISSUE #367: Define how to safety limits interact with clear history
It is a TODO in the spec to define how to safety limits interact with clear history.

### LINKED ISSUE #393: Clearing browsing history for any site effectively clears data for all sites
The `clear browsing history for attribution` [algorithm](https://w3c.github.io/attribution/#clear-browsing-history-for-attribution) takes a set of sites to support only clearing data for certain sites.

However, any call (with `forgetVisits` true) sets the `last browsing history clear`. While the note below the algorithm acknowledges that this "will make some impressions unreachable", it seems to make _all_ existing impressions unusable.

Is this the expected behavior? If so, should we consider updating that note and modifying the algorithm to just clear all three stores unconditionally?

martinthomson:
This is pretty straightforward once I thought it through.

Most things are left along if `forgetSites` is false, so that is super easy.

When it is true and the set of sites is empty, we already move the value of the last browsing history clear.  So we can just drop all state.  This is the easiest one.

When `forgetSites` is true and the set of sites is not empty, we do nothing special for the global privacy budget.  We can retain that. The impression site quota is also pretty easy.  We forget that site was ever there.  That would reset the amount that impressions from that site can eat from the global budget, except that the last browsing history clear value will ensure that none of those sites will be able to be found anyway.

This is the small change.  If you think about the effect of the last browing history clear value, it might be that the forgetting of specific sites - for this API - is effectively the same as forgetting all sites, so we can remove the distinction.  I guess that the reason we have the current structure is that we wanted to retain the possibility that impressions from non-forgotten sites would be reachable at some point. With the impression site quota limit in place, that might no longer be possible.  Still, baby steps.

Closes #367.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/pull/403.html" title="Last updated on Apr 23, 2026, 1:37 PM UTC (4e30a16)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/403/904ab26...4e30a16.html" title="Last updated on Apr 23, 2026, 1:37 PM UTC (4e30a16)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -1085,7 +1085,7 @@ that are used to manage the expenditure of [=privacy budgets=]:
     This value is initialized as a side effect
     of invoking <a method for=Attribution>measureConversion()</a>.
 
-*   A singleton [=last browsing history clear=] value
+*   A singular [=last browsing history clear=] value
     that tracks when the browsing activity for a [=site=] was last cleared.
 
 *   The [=global privacy budget store=] records the state
@@ -1409,7 +1409,7 @@ returning an [=epoch index=]:
     1.  Let |rand| be |t| minus a [=duration=]
         that is randomly selected
         from between 0 (inclusive) and |period| (exclusive).
-    
+
     1.  Let |ms| be the number of milliseconds in
         the [=duration from=] the [=unix epoch=] to |rand|.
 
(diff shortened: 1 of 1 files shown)


apasel422 (approved):
This also addresses #393, right?

api.bs
bmcase:
```suggestion
        Adding <em>one</em> is necessary so that the [=epoch=] range for attribution
```

bmcase:
>  If you think about the effect of the last browing history clear value, it might be that the forgetting of specific sites - for this API - is effectively the same as forgetting all sites, so we can remove the distinction. I guess that the reason we have the current structure is that we wanted to retain the possibility that impressions from non-forgotten sites would be reachable at some point.

Right, it seems currently because we move the last clear history we are essentially clearing for all sites (lot letting them query past impressions even if there sites was not in the set of sites to be cleared).   I guess if we wanted to allow that functionality we would need to keep not just one single clear history value but a site specific store of last history clears?

bmcase (approved):

martinthomson:
I've opened #404 to track a less disruptive per-site state clear.  That is going to be difficult.

PR #413, opened 2026-04-29
Define parallel activation for attribution

### LINKED ISSUE #372: Resetting the global attribution API flag
The spec has a TODO on defining how to reset the user activation flag.  

From @martinthomson  [comment ](https://github.com/w3c/attribution/pull/309#discussion_r2785594609)on this topic: 

> I think that we have reset conditions that come for free:
> 
> Navigation causes the value to clear, because that creates a new Window object.
> That is probably OK for now. What we might consider for reset include:
> 
> Extra long periods without interaction, such that the site execution is paused.
> What we might want to avoid:
> 
> Browser restarts. If those reactivate sessions, the site will expect the API to function.

### LINKED ISSUE #378: HTML's user activation modes don't work for us
A simple test confirms that our understanding of how activation works doesn't fit our requirement:

One page has a link:
```html
<a href="the-other.html">Click</a>
```

The other page reports whether it receives activation:
```html
<script>window.alert(`is ${navigator.userActivation.isActive} was ${navigator.userActivation.hasBeenActive}`)</script>
```

Spoiler: no activation on the second page, because activation is tracked on the `Window`. This should have been obvious to me (mea culpa), but I didn't read the spec closely enough.

Aside: Why browsers let sites use `window.alert()` without user activation is a thing from history: it should consume activation, but doing that would break things.

So, given that this doesn't work, what do we need to do is:

1. Track activation at the "[top-level traversable](https://html.spec.whatwg.org/#top-level-traversable)".  In simpler terms, this is a browser tab.
2. Replicate the [activation notification](https://html.spec.whatwg.org/#activation-notification) logic that the existing user activation logic does: On an [activation-triggering input event](https://html.spec.whatwg.org/#activation-triggering-input-event), set some state on the navigable.
3. Decide on what state we track.  If we want a single use of the API per activation, this can be a timestamp, just like for transient activation.  If we want more uses of the API, this needs to include a counter, or even a set that tracks which sites were able to use the API.

Then we need to use the information.  If it is just one use per activation, we can clear a timestamp, just like for transient activation.  Otherwise, it gets more complicated, but clearing the timestamp is probably a good part of any option.

We also need to decide how long to keep that state active, but I suggest that we rely on the browser's already-defined [transient activation duration](https://html.spec.whatwg.org/#transient-activation-duration).

---

The other part of this is to consider the role of [user navigation involvement)](https://html.spec.whatwg.org/#user-navigation-involvement).

I'm going to suggest that we consider this in our assessment.  If the site that receives activation is not the same as the current site, and [user navigation involvement](https://html.spec.whatwg.org/#user-navigation-involvement) is "[none](https://html.spec.whatwg.org/#uni-none)", then maybe we can throw.

The reason I say that is that navigation that is not the result of user action should not carry this state forward.  We can maybe track the site where the activation occurred and check that against the site where the API is invoked so that same-site automated navigations are OK.

However, for this part, I need to confirm that the navigation involvement is preserved across redirects.  More work is needed there.

Also, it's possible that I'm overthinking this part.

---

In the interest of making progress, I am going to suggest we move ahead with #309, acknowledging that this part is buggy in its current form.  That is, we don't insist that #309 "works" before landing it.  There are lots of other parts to that PR that are useful.

I'm going to hazard a guess and say that this fix doesn't require a bunch of discussion, because most of what we've been talking about was operating on the assumption that it is consistent with what we've discussed.

martinthomson:
This replicates much of the logic for transient activation, but rather than tracking the state on each Window, this tracks it in the top-level traversable (think browser tab).

The explanation is longer than the algorithms, but hopefully that means I don't need to repeat the rationale here.

I've gone with a bit of an unconventional method for monkey-patching the navigate algorithm.  I think that this is more robust than the normal method of identifying concrete steps.  I'd be interested in feedback on whether this is more comprehensible than other monkey-patching approaches. I despise monkey patching, but it seems unavoidable given the architecture of the specifications involved; hence my desire to find something less repugnant.

Closes #378.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/pull/413.html" title="Last updated on May 12, 2026, 2:21 PM UTC (4a44822)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/413/f37dcf5...4a44822.html" title="Last updated on May 12, 2026, 2:21 PM UTC (4a44822)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -1294,6 +1294,11 @@ in terms of how activation state is tracked.
 *   Any [=activation triggering input event|user activation=]
     makes the Attribution API available for a short period.
 
+*   Additionally, when [=navigate|navigation=] is initiated
+    by a user
+    (that is, where the [=user navigation involvement=] is "browser UI"),
+    the Attribution API is available for a short period.
+
 *   The first attempt to access the API by invoking
     either <a for=Attribution>saveImpression()</a>
     or <a for=Attribution>measureConversion()</a>
@@ -1365,13 +1370,24 @@ do not cause the Attribution API to become inaccessible.
 <div algorithm="attribution activation">
 When a user interaction causes firing of an [=activation triggering input event=]
 in a {{Document}} |document|,
-the user agent [=must=] perform the following steps--
+the user agent [=must=] perform the steps below--
 in addition to the [=activation notification=] steps--
-before dispatching the event:
+before dispatching the event.
+In this case, set [=navigable=] |navigable| to null.
 
-1.  Let |t| be the result of getting |document|'s [=node navigable=].
+Alternatively, when [=navigate|navigation starts=],
+and the [=user navigation involvement=] is `"browser UI"`,
+perform the steps below
+setting |navigable| is set to the affected [=navigable=]
+and setting |document| to null.
 
-1.  Let |top| be the result of getting |t|'s [=top-level traversable=].
+The steps are:
+
+1.  If |navigable| is not null, let |n| be |navigable|.
+
+1.  Otherwise, let |n| be the result of getting |document|'s [=node navigable=].
+
+1.  Let |top| be the result of getting |n|'s [=top-level traversable=].
 
 1.  Set |top|'s [=attribution activation timestamp=]
     to the [=current high resolution time=].
--- a/api.bs
+++ b/api.bs
@@ -1388,7 +1388,7 @@ In this case, set [=navigable=] |navigable| to null.
 Alternatively, when [=navigate|navigation starts=],
 and the [=user navigation involvement=] is `"browser UI"`,
 perform the steps below
-setting |navigable| is set to the affected [=navigable=]
+setting |navigable| to the affected [=navigable=]
 and setting |document| to null.
 
 The steps are:
--- a/api.bs
+++ b/api.bs
@@ -1413,9 +1413,9 @@ To <dfn>check attribution API activation</dfn>,
 given a {{Document}} |document|,
 throwing a {{"NotAllowedError"}} {{DOMException}} on failure:
 
-1.  Let |t| be the result of getting |document|'s [=node navigable=].
+1.  Let |n| be the result of getting |document|'s [=node navigable=].
 
-1.  Let |top| be the result of getting |t|'s [=top-level traversable=].
+1.  Let |top| be the result of getting |n|'s [=top-level traversable=].
 
 1.  If |top|'s [=attribution enabled flag=] is true, return.
 
--- a/api.bs
+++ b/api.bs
@@ -1316,7 +1316,7 @@ in terms of how activation state is tracked.
 
 This is similar in effect to [=transient activation=].
 However, attribution activation uses separate state from [=transient activation=]
-and has the following additional differences:
+and has the following notable differences:
 
 *   Attribution activation state is tracked on the [=top-level traversable=],
     rather than affected {{Window}}s.
--- a/api.bs
+++ b/api.bs
@@ -1367,6 +1367,16 @@ though a larger value
 might be advisable to ensure that delays from [=navigate|navigation=]
 do not cause the Attribution API to become inaccessible.
 
+<p class=note>
+Implementations might choose to extend the time allowed
+to account for the delays involved in page loading time.
+A fixed value might not account for performance differences
+in devices or in the network they use.
+Implementations could extend the value of [=attribution activation duration=]
+based on an understanding of performance chararacteristics.
+Alternatively, an implementation might suspend any timer
+during page load.
+
 <div algorithm="attribution activation">
 When a user interaction causes firing of an [=activation triggering input event=]
 in a {{Document}} |document|,
--- a/api.bs
+++ b/api.bs
@@ -1306,6 +1306,7 @@ in terms of how activation state is tracked.
 *   When activation is consumed,
     the API is enabled for use by all [=descendent navigables=]
     of the [=top-level traversable=]'s [=active document=]
+    (that is, all frames on the current page)
     until the next cross-site navigation.
 
 *   Any [=navigate|navigation=] of the [=top-level traversable=]
--- a/api.bs
+++ b/api.bs
@@ -1294,10 +1294,10 @@ in terms of how activation state is tracked.
 *   Any [=activation triggering input event|user activation=]
     makes the Attribution API available for a short period.
 
-*   The first attempt access the API by invoking
+*   The first attempt to access the API by invoking
     either <a for=Attribution>saveImpression()</a>
     or <a for=Attribution>measureConversion()</a>
-    within the an [=implementation-defined=] duration of the last activation,
+    within an [=implementation-defined=] duration of the last activation,
     enables the API.
 
 *   Like [=transient activation=],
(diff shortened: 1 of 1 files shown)


api.bs
```diff
@@ -1287,63 +1287,134 @@ integer |valueDeduction|,
...
+
+*   Any [=navigate|navigation=] of the [=top-level traversable=]
+    to a different [=site=] disables the API.
+    However, any [=activation triggering input event|activation=],
+    which might include the action that initiated the navigation,
+    could make the API available again.
+
+This is similar in effect to [=transient activation=].
+However, attribution activation uses separate state from [=transient activation=]
+and has the following additional differences:
```
```suggestion
and has the following notable differences:
```

api.bs
```diff
@@ -1287,63 +1287,134 @@ integer |valueDeduction|,
...
-while only making the API available to one site
-per <a href="https://html.spec.whatwg.org/#activation-triggering-input-event">activation</a>.
+The Attribution API is like a [=transient activation-consuming API=],
+with a number of differences
+in terms of how activation state is tracked.
+
+*   Any [=activation triggering input event|user activation=]
+    makes the Attribution API available for a short period.
+
+*   The first attempt access the API by invoking
```
```suggestion
*   The first attempt to access the API by invoking
```

api.bs
```diff
@@ -1287,63 +1287,134 @@ integer |valueDeduction|,
...
+    either <a for=Attribution>saveImpression()</a>
+    or <a for=Attribution>measureConversion()</a>
+    within the an [=implementation-defined=] duration of the last activation,
```
```suggestion
    within an [=implementation-defined=] duration of the last activation,
```

api.bs
```diff
@@ -1287,63 +1287,134 @@ integer |valueDeduction|,
...
+1.  The <dfn>attribution enabled flag</dfn> is a boolean value
+    that tracks whether the API is enabled.
+    This value is initialized to false.
+
+1.  The <dfn>attribution activation timestamp</dfn> is a [=moment=]
+    that tracks the last [=activation triggering input event|user activation=].
+    This value is initialized to the [=current high resolution time=].
+
+Implementations also configure <dfn>attribution activation duration</dfn>
+as an an [=implementation-defined=] [=duration=].
```
```suggestion
as an [=implementation-defined=] [=duration=].
```

api.bs
```diff
@@ -1287,63 +1287,134 @@ integer |valueDeduction|,
...
+This should be no less than the [=transient activation duration=],
+though a larger value
+might be advisable to ensure that delays from [=navigate|navigation=]
+do not cause the attribution API to become inaccessible.
```
```suggestion
do not cause the Attribution API to become inaccessible.
```

api.bs
```diff
@@ -1287,63 +1287,134 @@ integer |valueDeduction|,
...
 
-    1.  [=Consume user activation=] given |window|.
+1.  If |top|'s [=attribution enabled flag=] is true, return.
 
-    1.  Set |window|'s [=global attribution API flag=] to true.
+1.  Let |lastActivation| be |top|'s [=attribution activation timestamp=].
 
-    1.  Return.
+1.  If |lastActivation| plus the [=attribution activation duration=]
+    is less than the [=current high resolution time=],
```
apasel422, line 1435: Extremely minor, but do we want to reuse the impression/conversion time from the implicit API inputs, rather than re-obtain the time? If so, we can make this algorithm accept the implicit inputs as a parameter, since they already contain the document.

apasel422 (changes requested):

api.bs
martinthomson: I considered that.  It would be a sensible thing to do if you had a real implementation, not least because you don't want to poll the clock more often than you need to.  But there is also value in this being somewhat disconnected from the rest of the API.  I settled on this for simplicity of one sort, but if you think it's worth the change, I can change to that approach.  Or I could have the time passed in, maintaining modularity or distance from the implicit inputs, but avoiding extra hits on the clock.  That seemed like an implementation optimization though.  ㄟ( ▔, ▔ )ㄏ

apasel422: Sounds fine for now.

apasel422 (approved):
LGTM but @bmcase should review too.

api.bs
```diff
@@ -1287,63 +1287,134 @@ integer |valueDeduction|,
...
+1.  The <dfn>attribution activation timestamp</dfn> is a [=moment=]
+    that tracks the last [=activation triggering input event|user activation=].
+    This value is initialized to the [=current high resolution time=].
+
+Implementations also configure <dfn>attribution activation duration</dfn>
+as an [=implementation-defined=] [=duration=].
+This should be no less than the [=transient activation duration=],
+though a larger value
+might be advisable to ensure that delays from [=navigate|navigation=]
+do not cause the Attribution API to become inaccessible.
```
bmcase, line 1368: I think when we had discussed and drafted something for user activation before, it was that case that when there was a user initiated navigation to a site the API would remain activated for that site at least for that session (with some site level enabled flag).  Is that changing if the duration is the same as the transient activation duration which is only a few seconds?

bmcase (approved):
left a comment we should discuss and/or follow up on to make sure this can support the functionality of a user being on a site for a while and the API being called over time and not just when the user first navigates to the site.

api.bs
```diff
@@ -1287,63 +1287,134 @@ integer |valueDeduction|,
...
+in terms of how activation state is tracked.
+
+*   Any [=activation triggering input event|user activation=]
+    makes the Attribution API available for a short period.
+
+*   The first attempt to access the API by invoking
+    either <a for=Attribution>saveImpression()</a>
+    or <a for=Attribution>measureConversion()</a>
+    within an [=implementation-defined=] duration of the last activation,
+    enables the API.
```
martinthomson, line 1306:
```suggestion
    enables the API for the current top-level site.
```

api.bs
```diff
@@ -1287,63 +1287,134 @@ integer |valueDeduction|,
...
+
+*   Like [=transient activation=],
+    using the API consumes the activation.
+
+*   When activation is consumed,
+    the API is enabled for use by all [=descendent navigables=]
+    of the [=top-level traversable=]'s [=active document=]
+    until the next cross-site navigation.
```
martinthomson, line 1315:
```suggestion
    (that is, all frames on the current page)
    until the next cross-site navigation.
```

api.bs
```diff
@@ -1287,63 +1287,134 @@ integer |valueDeduction|,
...
+and has the following notable differences:
+
+*   Attribution activation state is tracked on the [=top-level traversable=],
+    rather than affected {{Window}}s.
+
+*   Attribution activation persists through [=navigate|navigations=],
+    enabling use of the API for a short period
+    after loading a new page
+    in the case where the activation leads to navigation.
+
```
martinthomson, line 1334: We should consider whether a user that enters a fresh URI in the address bar clears the activation status.  Or whether this results in the activation being engaged when that page first loads.

api.bs
martinthomson, line 1368: It might be worth an implementation note here: the timer could be extended by the time it takes to load pages during navigation.  Effectively pausing the timer while no page is actively running.

bmcase: notes:  discussed that this still supports this functionality in that the algorithm will still have a flag to enable the API for the site and once it is set it will remain.  So as long as a site activates the API in the first few seconds, it will remain active until the user navigates away.

bmcase: also if a user continues to click on that site, such as to other pages, there will be further opportunities to activate the API

api.bs
```diff
@@ -1365,13 +1370,24 @@ do not cause the Attribution API to become inaccessible.
...
 in addition to the [=activation notification=] steps--
-before dispatching the event:
+before dispatching the event.
+In this case, set [=navigable=] |navigable| to null.
 
-1.  Let |t| be the result of getting |document|'s [=node navigable=].
+Alternatively, when [=navigate|navigation starts=],
+and the [=user navigation involvement=] is `"browser UI"`,
+perform the steps below
+setting |navigable| is set to the affected [=navigable=]
```
bmcase:
```suggestion
setting |navigable| to the affected [=navigable=]
```

bmcase:
looks good to support activation on user navigation through browser UI.

PR #420, opened 2026-05-12
Add activation to simulator

apasel422:


geejlyn76-max (approved):

impl/e2e-tests/activation.json
```diff
@@ -0,0 +1,61 @@
...
+      "event": "updateActivationTimestamp",
+      "topLevelTraversable": "a"
+    },
+    {
+      "seconds": 1002,
+      "site": "publisher.example",
+      "event": "saveImpression",
+      "topLevelTraversable": "a",
+      "options": { "histogramIndex": 1 },
+      "expectedError": {
```
martinthomson, line 14: expectedError here, expected below ??!?

impl/e2e-tests/activation.json
```diff
@@ -0,0 +1,61 @@
...
+        "name": "NotAllowedError"
+      }
+    },
+    {
+      "seconds": 1004,
+      "event": "updateActivationTimestamp",
+      "topLevelTraversable": "b"
+    },
+    {
+      "seconds": 2004,
```
martinthomson, line 41: This works because the time is *exactly* 1000 seconds, yes?  And the conversion measurement works at 1001 seconds because of the flag.

impl/e2e-tests/activation.json
```diff
@@ -0,0 +1,61 @@
...
+      "event": "measureConversion",
+      "topLevelTraversable": "b",
+      "options": {
+        "aggregationService": "https://agg-service.example",
+        "histogramSize": 3,
+        "value": 5,
+        "maxValue": 10
+      },
+      "expected": [0, 5, 0]
+    }
```
martinthomson, line 59: I would add a cross-site navigation on "b" here to demonstrate that the errors start back up.

martinthomson (changes requested):
Approval pending, but I think that you need to use the navigation hook before this is done.

impl/e2e-tests/activation.json
apasel422: They are different for `saveImpression` and `measureConversion`, as the latter can expect either an error or a value.

impl/e2e-tests/activation.json
geejlyn76-max: > I would add a cross-site navigation on "b" here to demonstrate that the errors start back up.

PR #431, opened 2026-05-13
Some editorial improvements to activation text

### LINKED ISSUE #430: Clarifications on activation logic
I've been looking at [Section 4.2.2](https://w3c.github.io/attribution/#s-api-activation) as part of #420 and want to clarify two things:

First, the prose says that a top-level traversable's `attribution activation timestamp` is "initialized to the current high resolution time."

I read this to mean that when a top-level traversable is constructed (regardless of how), that field is initialized that way, meaning that there is a (probably brief) period of time in which the traversable has activation, allowing immediate `saveImpression` or `measureConversion` calls to succeed without any explicit user interaction or "browser UI" navigation involvement.

Is this intended, or should the initial state be null or `MIN_TIME` to ensure that the very first access always requires a deliberate user action?

Second, the prose says that "using the API consumes the activation," drawing a parallel to transient activation. However, the [`check attribution API activation` algorithm](https://w3c.github.io/attribution/#check-attribution-api-activation) only sets the `attribution enabled flag` to `true` and does not modify the `attribution activation timestamp`, so I'm not really sure what is being consumed here, unless the algorithm is missing a step to clear/reset the `attribution activation timestamp`? If the algorithm is correct, maybe we can avoid using the term `consume` to avoid confusion with the concept from transient activation.

martinthomson:
... and a change to the initialization value for the activation timestamp.

Closes #430.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/pull/431.html" title="Last updated on May 14, 2026, 7:35 PM UTC (d1a651b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/431/28b5c09...d1a651b.html" title="Last updated on May 14, 2026, 7:35 PM UTC (d1a651b)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -1300,9 +1300,11 @@ integer |valueDeduction|,
 
 ### Attribution API Activation ### {#s-api-activation}
 
-The Attribution API is like a [=transient activation-consuming API=],
-with a number of differences
-in terms of how activation state is tracked.
+The Attribution API is roughly modeled on [=transient activation-consuming APIs=],
+where [=activation triggering input event|user activation=]
+is temporarily makes the API available.
+A number of differences distinguish this design
+from [=transient activation-consuming APIs=]:
 
 *   Any [=activation triggering input event|user activation=]
     makes the Attribution API available for a short period.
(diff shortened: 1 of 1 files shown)


api.bs
```diff
@@ -1300,9 +1300,11 @@ integer |valueDeduction|,
 
 ### Attribution API Activation ### {#s-api-activation}
 
-The Attribution API is like a [=transient activation-consuming API=],
-with a number of differences
-in terms of how activation state is tracked.
+The Attribution API is roughly modeled on [=transient activation-consuming APIs=],
+where [=activation triggering input event|user activation=]
+is temporarily makes the API available.
```
```suggestion
temporarily makes the API available.
```

apasel422 (approved):

PR #439, opened 2026-05-26
Credit needs to be finite

martinthomson:
We don't specify a maximum value in the same way we do for epsilon, so this would have accepted `Infinity`, which isn't exactly going to work out well.  Now, `[Number.MAX_VALUE, Number.MAX_VALUE]` still leads to a bad outcome, but trying to prevent that sort of nonsense isn't something I think we need to over-index on. 

The implementation already checks that the value is finite; this just has the spec catch up.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/pull/439.html" title="Last updated on May 26, 2026, 1:17 AM UTC (731322e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/439/b2461cc...731322e.html" title="Last updated on May 26, 2026, 1:17 AM UTC (731322e)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -1929,7 +1929,10 @@ To <dfn>validate {{AttributionConversionOptions}}</dfn> |options|:
     throw a {{RangeError}}.
 1.  Let |credit| be |options|.{{AttributionConversionOptions/credit}} if it [=map/exists=], «1» otherwise.
 1.  If |credit| [=list/is empty=], throw a {{RangeError}}.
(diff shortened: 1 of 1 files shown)


apasel422:
There's no harm in adding this check, but IIUC the [WebIDL double](https://webidl.spec.whatwg.org/#idl-double) type is already required to be finite, which is why [the simulator uses an assertion for this](https://github.com/w3c/attribution/blob/b2461cc2e9043d87528267c9d04275d7f792caf5/impl/src/backend.ts#L276).

bmcase (approved):

PR #441, opened 2026-05-29
Limitations section

martinthomson:
There is a lot of this document that talks about what it can do, but that fails to account for potential misapprehension about what is possible.

This section attempts to enumerate limitations when it comes to using this API for the measurement of advertising effectiveness, particularly when it comes to producing information that is helpful in making decisions about where to invest in marketing.

I've put this up front, so the disclaimer is clear.  The section is longer than many of the adjoining sections; I hope that conveys the right sort of message.

Thanks to @rickcentralcontrolcom for raising the underlying issue.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/pull/441.html" title="Last updated on Jun 18, 2026, 12:00 AM UTC (0e96ada)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/441/e3f147d...0e96ada.html" title="Last updated on Jun 18, 2026, 12:00 AM UTC (0e96ada)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -289,80 +289,86 @@ However, without due care, use of the API
 could produce misleading results that support poor decisions.
 
 To make better decisions using the information provided by the API,
-its limitations
-especially the potential sources of bias and error,
+its limitations--
+especially the potential sources of bias and error--
 need to be understood.
 Understanding needs to be applied
 both in the design of systems that use the API
 and in the interpretation of any results.
 
-**Scope limitations**--
-A critical limitation is that only activity in the same browser
-can be captured by the Attribution API.
-The extent and effects of advertising campaigns
-are rarely contained to a single medium.
-This means that any measurement results obtained
-cannot always capture all relevant activity.
+<dl>
 
-**Non-causal effects**--
-Some proportion of people who encounter advertising
-already intend to follow through with the actions
-the advertising seeks to encourage.
-Attributing results with those advertisements
-creates a false impression of their efficacy.
+: Scope limitations
+::  A critical limitation is that only activity in the same browser
+    can be captured by the Attribution API.
+    The extent and effects of advertising campaigns
+    are rarely contained to a single medium.
+    This means that any measurement obtained
+    cannot be assumed to capture all relevant activity.
 
-To better measure the causal effect of advertising,
-the use of (randomized) control trials--
-or, in industry terms, incrementality--
-is necessary [[JOHNSON20]].
-A basic advertising control trial
-involves identifying a target audience,
-advertising to a treatment group within that audience,
-withholding advertising or showing default advertisements to a control group,
-and measuring the difference in desired outcomes between the groups.
+: Non-causal effects
+::  Some proportion of people who encounter advertising
+    already intend to follow through with the actions
+    the advertising seeks to encourage.
+    Attributing outcomes to those advertisements
+    creates a false impression of their efficacy.
 
-This API can be used to support such trials.
-This might be achieved by allocating control and treatment populations
-to different histogram buckets.
-Note however that the web platform presently does not have a facility for
-providing consistent allocation of users
-into control and treatment groups
-across different sites.
+::  To better measure the causal effect of advertising,
+    the use of (randomized) control trials--
+    or, in industry terms, incrementality--
+    is necessary [[INFERNO]].
+    A basic advertising control trial
+    involves identifying a target audience,
+    advertising to a treatment group within that audience,
+    withholding advertising or showing default advertisements to a control group,
+    and measuring the difference in desired outcomes between the groups.
 
-**Simple attribution logic**--
-The API provides a simple multi-touch attribution algorithm.
-Simple attribution algorithms can introduce bias into results
-in ways that effect decision-making.
-Future versions of the API
-might enable more sophisticated algorithms.
+::  This API can be used to support such trials.
+    This might be achieved by allocating control and treatment populations
+    to different histogram buckets.
+    Note however that the web platform presently does not have a facility for
+    providing consistent allocation of users
+    into control and treatment groups
+    across different sites.
 
-**Sources of error**--
-There are multiple sources of error
-in a distributed measurements system.
-Differential privacy contributes some amount of noise to aggregates,
-but this is a known quantity.
+: Simple attribution logic
+::  The API provides a simple multi-touch attribution algorithm.
+    Simple attribution algorithms can introduce bias into results
+    in ways that affect decision-making.
+    Future versions of the API
+    might enable more sophisticated algorithms.
 
-Results might not accurately reflect what is being measured
-for a range of other reasons,
-some of which might be unavoidable,
-including:
+: Sources of error
+::  There are multiple sources of error
+    in a distributed measurement system.
+    Differential privacy contributes some amount of noise to aggregates,
+    but this is a known quantity.
 
-*   Failures and delays in propagating configuration to participating sites.
-*   Script errors that cause loss of impression or conversion events.
-*   Loss of stored impressions from storage pressure.
-*   Accidential or malicious recording of impressions.
-*   Network errors that result in lost reports.
-*   <abbr title="InValid Traffic">IVT</abbr> classification that either
-    causes genuine events to be disregarded
-    or includes fraud and accidental mistakes.
+::   Results might not accurately reflect what is being measured
+    for a range of other reasons,
+    some of which might be unavoidable,
+    including:
 
-Errors such as these could add an unknown amount of uncertainty
-to measurement results.
-Sites that use the API are responsible
-for managing their exposure to these and other classes of error.
-Future versions of the API
-could seek to provide better assistance to sites
-in handling errors.
+    *   Failures and delays in propagating configuration to participating sites.
+    *   Script errors that cause loss of impression or conversion events.
+    *   Loss of stored impressions from storage pressure.
+    *   Accidental or malicious recording of impressions.
+    *   Network errors that result in lost reports.
+    *   <abbr title="invalid traffic">IVT</abbr> classification that
+        causes genuine events to be disregarded.
+    *   <abbr title="invalid traffic">IVT</abbr> classification that
+        erroneously includes fraud or accidental mistakes.
+
+
+::  Errors such as these could add an unknown amount of uncertainty
+    to measurement results.
+    Sites that use the API are responsible
+    for managing their exposure to these and other classes of error.
+    Future versions of the API
+    could seek to provide better assistance to sites
+    in handling errors.
+
+</dl>
 
 
 # Overview of Operation # {#overview}
@@ -4069,7 +4075,7 @@ spec:css-2025; type:dfn; text:user
     "publisher": "Bureau of Economic Analysis",
     "date": "2017-10"
   },
-  "johnson20": {
+  "inferno": {
     "authors": [
       "Garrett A. Johnson"
     ],
--- a/api.bs
+++ b/api.bs
@@ -12,7 +12,15 @@ Editor: Benjamin Case, w3cid 128082, Meta https://www.meta.com/, bmcase@meta.com
 Editor: Benjamin Savage, w3cid 114877, Meta https://www.meta.com/, btsavage@meta.com
 Editor: Charlie Harrison, w3cid 110615, Google https://www.google.com/, csharrison@chromium.org
 Editor: Martin Thomson, w3cid 68503, Mozilla https://mozilla.org/, mt@mozilla.com
-Abstract: This specifies a browser API for the measurement of advertising performance.  The goal is to produce aggregate statistics about how advertising leads to conversions, without creating a risk to the privacy of individual web users.  This API collates information about people from multiple web origins, which could be a significant risk to their privacy.  To manage this risk, the information that is gathered is aggregated using an aggregation service that is trusted by the user-agent to perform aggregation within strict limits.  Noise is added to the aggregates produced by this service to provide differential privacy. Websites may select an aggregation service from the list of approved aggregation services provided by the user-agent.
+Abstract:
+  This specifies a browser API for attribution.
+  The API produces aggregate statistics that can help sites
+  better understand the real-world performance of their advertising.
+  This API collates information about people from multiple web origins,
+  which could be a significant risk to their privacy.
+  To manage this risk,
+  the information that is gathered is aggregated
+  and differential privacy techniques are applied.
 Complain About: accidental-2119 yes, missing-example-ids yes
 Markup Shorthands: markdown yes, css no, dfn yes
 Assume Explicit For: yes
@@ -24,9 +32,12 @@ Level: None
 # Introduction # {#intro}
 
 This document defines an API for browsers
-that enables the collection of aggregated, differentially-private metrics.
+that enables the collection of [[#aggregation|aggregated]],
+[[#dp|differentially-private]] cross-site metrics.
 
 The goal of this API is to enable attribution for advertising.
+These aggregated metrics can contribute to a better understanding of advertising performance,
+provided that API is [[#limitations|used appropriately]].
 
 
 ## Attribution ## {#s-attribution}
@@ -122,9 +133,16 @@ with multiple purveyors of personal information that is traded for various purpo
 
 ## Goals ## {#goals}
 
-The goal of this document is to define a means of performing [=attribution=]
-for advertising
-that does not enable tracking.
+This document seeks to enable [=attribution=]
+for sites that use advertising
+while avoiding mechanisms that enable or encourage [[UNSANCTIONED-TRACKING|tracking]],
+such as [[WEB-WITHOUT-3P-COOKIES|cross-site cookies]].
+
+The Attribution API aims to provide
+[[#user-benefit|indirect]], [[#collective|collective]] user benefit
+by giving sites measurement tools.
+These tools can provide insight into advertising performance
+when [[#limitations|applied effectively]].
 
 
 ## End-User Benefit ## {#user-benefit}
@@ -155,19 +173,20 @@ path:images/value.svg
 Participation in an [=attribution=] measurement system
 would comprise a secondary cost to Web users.
 
-Support for attribution enables more effective advertising,
-largely by helping advertisers understand which ads perform best,
-and in what circumstances.
-Those circumstances might include
-the time and place that the ad is shown,
+Support for attribution can enables more effective advertising,
+by giving advertisers information
+they can use to improve advertising strategies.
+Attribution enables measurement of correlations
+between desired outcomes with characteristics of ads,
+such as the time and place that the ad is shown,
 the person to whom the ad is presented, and
 the details of the ad itself.
+Sites can use this information to improve how they use advertising.
 
-Connecting that information to outcomes
-allows an advertiser to learn what circumstances most often lead
-to the outcomes they most value.
-When [[#limitations|attribution is used effectively]],
-it allows advertisers to spend more on effective advertising
+Connecting actions to outcomes
+is a tool that advertisers can use
+as part of a larger measurement strategy that improves performance.
+Good measurement can allow advertisers to spend more on effective advertising
 and less on ineffective advertising.
 This lowers the overall cost of advertising
 relative to the value obtained. [[ONLINE-ADVERTISING]]
@@ -266,7 +285,7 @@ This enables both
 rigorous [[#limitations|experiments that seek to establish causality]]
 and simpler comparisons of different treatments.
 For instance, grouping by creative (the content of an ad)
-might be used to learn which creative works best.
+might be used to compare different creatives.
 
 Adding a value greater than one at each conversion
 enables more than simple counts.
--- a/api.bs
+++ b/api.bs
@@ -173,7 +173,7 @@ path:images/value.svg
 Participation in an [=attribution=] measurement system
 would comprise a secondary cost to Web users.
 
-Support for attribution can enables more effective advertising,
+Support for attribution can enable more effective advertising,
 by giving advertisers information
 they can use to improve advertising strategies.
 Attribution enables measurement of correlations
--- a/api.bs
+++ b/api.bs
@@ -281,6 +281,90 @@ to split credit.
 -->
 
 
+## Limitations and Successful Use ## {#limitations}
+
+The results that the Attribution API provides
+can inform decisions about advertising strategies.
+However, without due care, use of the API
+could produce misleading results that support poor decisions.
+
+To make better decisions using the information provided by the API,
+its limitations
+especially the potential sources of bias and error,
+need to be understood.
+Understanding needs to be applied
+both in the design of systems that use the API
+and in the interpretation of any results.
+
+**Scope limitations**--
+A critical limitation is that only activity in the same browser
+can be captured by the Attribution API.
+The extent and effects of advertising campaigns
+are rarely contained to a single medium.
+This means that any measurement results obtained
+cannot always capture all relevant activity.
+
+**Non-causal effects**--
+Some proportion of people who encounter advertising
+already intend to follow through with the actions
+the advertising seeks to encourage.
+Attributing results with those advertisements
+creates a false impression of their efficacy.
+
+To better measure the causal effect of advertising,
+the use of (randomized) control trials--
+or, in industry terms, incrementality--
+is necessary [[JOHNSON20]].
+A basic advertising control trial
+involves identifying a target audience,
+advertising to a treatment group within that audience,
+withholding advertising or showing default advertisements to a control group,
+and measuring the difference in desired outcomes between the groups.
+
+This API can be used to support such trials.
+This might be achieved by allocating control and treatment populations
+to different histogram buckets.
+Note however that the web platform presently does not have a facility for
+providing consistent allocation of users
+into control and treatment groups
+across different sites.
+
+**Simple attribution logic**--
+The API provides a simple multi-touch attribution algorithm.
+Simple attribution algorithms can introduce bias into results
+in ways that effect decision-making.
+Future versions of the API
+might enable more sophisticated algorithms.
+
+**Sources of error**--
+There are multiple sources of error
+in a distributed measurements system.
+Differential privacy contributes some amount of noise to aggregates,
+but this is a known quantity.
+
+Results might not accurately reflect what is being measured
+for a range of other reasons,
+some of which might be unavoidable,
+including:
+
+*   Failures and delays in propagating configuration to participating sites.
+*   Script errors that cause loss of impression or conversion events.
+*   Loss of stored impressions from storage pressure.
+*   Accidential or malicious recording of impressions.
+*   Network errors that result in lost reports.
+*   <abbr title="InValid Traffic">IVT</abbr> classification that either
+    causes genuine events to be disregarded
+    or includes fraud and accidental mistakes.
+
+Errors such as these could add an unknown amount of uncertainty
+to measurement results.
+Sites that use the API are responsible
+for managing their exposure to these and other classes of error.
+Future versions of the API
+could seek to provide better assistance to sites
+in handling errors.
+
+
 # Overview of Operation # {#overview}
 
 The Attribution API provides aggregate information about the
@@ -3984,6 +4068,15 @@ spec:css-2025; type:dfn; text:user
     "href": "https://www.bea.gov/research/papers/2017/measuring-free-digital-economy-within-gdp-and-productivity-accounts",
     "publisher": "Bureau of Economic Analysis",
     "date": "2017-10"
+  },
+  "johnson20": {
+    "authors": [
+      "Garrett A. Johnson"
+    ],
+    "title": "Inferno: A Guide to Field Experiments in Online Display Advertising",
+    "href": "https://papers.ssrn.com/sol3/papers.cfm?abstract_id=3581396",
+    "publisher": "Journal of Economics & Management Strategy",
+    "date": "2023-01-18"
   },
   "online-advertising": {
     "authors": [
--- a/api.bs
+++ b/api.bs
@@ -98,7 +98,7 @@ The cost of measurement performance was privacy.
 In order to produce accurate and comprehensive information,
 advertising businesses performed extensive tracking of the activity of all Web users.
 Each browser was given a tracking identifier,
-often using cookies that were logged by cross-site content.
+often using cookies that associated with cross-site resources.
 Every action of interest was logged against this identifier,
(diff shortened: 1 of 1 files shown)


csharrison (approved):

bmcase (approved):

rickcentralcontrolcom:
Hello Martin,

Thank you for the note, and for taking the concern seriously.

I have compared the currently posted 29 May draft (
https://www.w3.org/TR/attribution/) against a copy I saved from 25 May. I
may be missing something, but I am unable to identify changes that
materially address the issue I raised. The sentence in §1.6 stating that
attribution “attempts to measure correlation” appears in both versions. If
there is a new or moved section that I should be looking at, please point
me to it.

My concern is not that the draft lacks any caveat. It is that the current
draft still repeatedly characterizes attribution as a way to identify
advertising effectiveness, when the API as described produces aggregate
information about associations between impressions and conversions. That
distinction is fundamental.
Before turning to specific passages, I want to restate the underlying issue
because I think it risks getting lost in discussions about individual
wording choices.

The central problem is the longstanding and well-established distinction
between correlation and causation.

"Correlation does not imply causation" is among the most fundamental
principles in statistics, economics, epidemiology, and scientific inference
generally. It is often taught as a simple mnemonic because humans are
naturally inclined to infer causality from observed associations. Yet that
inference is frequently wrong.

Roosters crow shortly before sunrise every day, but they do not cause the
sun to rise. Ice cream sales increase during summer months, and deaths by
drowning also increase during summer months, but neither causes the other.
Both are driven by a third factor. The existence of a statistical
relationship, even a strong one, does not establish that one event caused
another.

This confusion is hardly unique to this document. It has been a recurring
challenge throughout the history of advertising measurement. Attribution
systems observe exposures and subsequent outcomes, then identify
statistical associations between them. That does not mean the advertising
caused the outcome. Consumers already inclined to purchase are more likely
to search, click, visit websites, engage with advertising, and ultimately
convert. Distinguishing persuasion from pre-existing purchase intent is
precisely the problem that causal inference methodologies seek to solve.

Part of the difficulty is that causal assumptions are subtly embedded
throughout ordinary language. Words and phrases such as:

• effective advertising
• ineffective advertising
• perform best
• works best
• leads to
• drives
• causes
• influence
• impact
• effect
• improves advertising performance
• enables more effective advertising
• what works
• what ads perform best
• advertising effectiveness
• return on advertising
• successful advertising

all carry an implicit causal meaning. Readers naturally interpret these
phrases as statements about what advertising actually changed, not merely
what advertising happened to be associated with.

For that reason, I believe the Working Group should systematically review
the document for language that implicitly converts correlation into
causation. The issue is not confined to any single paragraph. It appears
throughout the narrative framing of the specification.

This matters because Attribution Level 1 is not merely a technical
proposal. It is a proposed web standard that could shape industry practice
for years. It carries the imprimatur of major standards and industry
organizations, including W3C, and will inevitably be cited by vendors,
platforms, agencies, consultants, publishers, and advertisers as an
authoritative description of what attribution measurement is capable of
determining.

The Working Group has understandably devoted enormous care to the technical
design, privacy safeguards, aggregation mechanisms, and implementation
details of the API. I believe the same level of care is required in
describing what the resulting measurements actually mean.

If the specification repeatedly characterizes attribution as a means of
determining advertising effectiveness, identifying what advertising works,
or distinguishing effective from ineffective advertising, many readers will
reasonably conclude that the standard itself endorses those claims. In my
view, that would be a serious mistake.

The consequences are not merely academic. Advertising measurement
frameworks influence the allocation of hundreds of billions of dollars
annually. When correlation is mistaken for causation, budgets tend to
migrate toward channels that are especially good at capturing existing
demand rather than creating incremental demand. Search, social, and retail
media frequently benefit from this dynamic because they operate closest to
observable conversion activity and possess extensive behavioral, identity,
and optimization capabilities.

Meanwhile, media channels whose effects are often delayed, indirect,
probabilistic, or difficult to observe through attribution
systems—including television, audio, premium video, sponsorships, and
broader brand advertising—tend to be systematically undercredited.

The result is not simply measurement error. It is the potential
misallocation of billions of dollars of advertising investment and further
economic pressure on media businesses that are already struggling to
survive. A standards document should therefore be especially careful not to
elevate an observational attribution framework into an implied measure of
causal advertising effectiveness.
The following passages in the 29 May draft still seem problematic.

In the Abstract:

Current language:
“This specifies a browser API for the measurement of advertising
performance. The goal is to produce aggregate statistics about how
advertising leads to conversions...”

Concern:
“Advertising performance” and “how advertising leads to conversions” both
suggest causal measurement.

Suggested revision:
“This specifies a browser API for privacy-preserving aggregate attribution
reporting. The goal is to produce aggregate statistics about associations
between advertising-related events and subsequent conversion events...”

In §1.2 Background:

Current language:
“One characteristic that distinguished the Web from other venues for
advertising was the ability to obtain information about the effectiveness
of advertising campaigns.”

Concern:
This frames attribution-era web measurement as effectiveness measurement,
rather than exposure, conversion, and association reporting.

Suggested revision:
“One characteristic that distinguished the Web from other venues for
advertising was the ability to obtain timely information about ad
exposures, interactions, and subsequent conversion events.”

In §1.2 Background:

Current language:
“Having a detailed record of a person’s actions allowed advertisers to
infer characteristics about people. Those characteristics made it easier to
choose the right audience for advertising, greatly improving its
effectiveness.”

Concern:
This again asserts improved effectiveness without distinguishing targeting
efficiency, observed conversion rates, and incremental causal impact.

Suggested revision:
“Those characteristics made it easier to target audiences believed to be
more likely to convert.”

In §1.2 Background:

Current language:
“Advertisers seek to place advertising where it will have the most effect
relative to its cost.”

Concern:
This is a reasonable business objective, but in context it implies that
attribution identifies where advertising has the most effect.

Suggested revision:
“Advertisers seek to place advertising where they expect the greatest
return relative to cost.”

In §1.3 Goals / §1.4 transition:

Current language:
“The measurement of advertising performance creates new cross-site flows of
information.”

Concern:
Again, “advertising performance” is broader than what the API establishes.

Suggested revision:
“Attribution reporting creates new cross-site flows of information.”

In §1.4 End-User Benefit:

Current language:
“Support for attribution enables more effective advertising, largely by
informing advertisers about what ads perform best, and in what
circumstances.”

Concern:
This is the clearest overstatement. Attribution can identify which ads are
associated with observed outcomes. It does not generally identify which ads
caused incremental outcomes or “perform best.”

Suggested revision:
“Support for attribution provides advertisers with aggregate information
about which ads, users, contexts, or circumstances are associated with
observed conversion events.”

In §1.4 End-User Benefit:

Current language:
“Connecting that information to outcomes allows an advertiser to learn what
circumstances most often lead to the outcomes they most value.”

Concern:
“Lead to” implies causation.

Suggested revision:
“Connecting that information to outcomes allows an advertiser to observe
which circumstances are most often associated with the outcomes they value.”

In §1.4 End-User Benefit:

Current language:
“That allows advertisers to spend more on effective advertising and less on
ineffective advertising.”

Concern:
This directly claims that attribution identifies effective and ineffective
advertising. That is the core issue.

Suggested revision:
“Advertisers may use this information, together with other measurement
methods, to inform campaign diagnostics, optimization, and investment
decisions.”

In §1.4 End-User Benefit:

Current language:
“Sites that provide advertising inventory... indirectly benefit from more
efficient advertising. Venues for advertising that are better able to show
ads that result in the outcomes that advertisers seek can charge more for
ad placements.”

Concern:
This again treats attributed outcomes as evidence of advertising efficiency
or causal contribution.

Suggested revision:
“Sites that provide advertising inventory may indirectly benefit when
advertisers have better aggregate reporting about observed outcomes
associated with ad placements.”

In §1.6 Attribution Using Histograms:

Current language:
“Different groupings might be used for different purposes. For instance,
grouping by creative (the content of an ad) might be used to learn which
creative works best.”

Concern:
“Works best” implies causal creative effectiveness. A histogram can compare
attributed outcomes by creative, but not necessarily incremental creative
effect.

Suggested revision:
“Different groupings might be used for different purposes. For instance,
grouping by creative might be used to compare observed attributed outcomes
across creatives.”

In §2 Overview of Operation:

Current language:
“Not displaying an advertisement (especially for controlled experiments
that seek to confirm whether an advertising campaign is effective).”

Concern:
This is less problematic because it refers to controlled experiments, but
it should be clearer that experimental designs are distinct from ordinary
attribution reporting.

Suggested revision:
“Not displaying an advertisement, where the API is used as part of a
controlled experimental design intended to estimate incremental advertising
effects.”

In §3.5 Requesting Attribution for a Conversion:

Current language:
“A site that observes a conversion might choose to request the measurement
of the effect of different stored impressions.”

Concern:
This is another causal claim. The API allocates conversion value to stored
impressions according to attribution logic. It does not measure the causal
effect of those impressions.

Suggested revision:
“A site that observes a conversion might choose to request attribution
reporting across different stored impressions.”

I would recommend adding a clear statement near the front of the document,
perhaps in §1.1 or §1.3, along these lines:

“This API produces aggregate attribution reports describing associations
between recorded impression events and subsequent conversion events. The
API does not, by itself, estimate the causal or incremental effect of
advertising. Claims about advertising effectiveness, incrementality, or
causal lift require additional methodological assumptions or separate
experimental or causal-inference designs.”

That clarification would be much stronger than a passing reference to
correlation in §1.6, especially while the rest of the draft continues to
use terms such as “effective advertising,” “perform best,” “works best,”
and “measurement of the effect.”

I am not objecting to privacy-preserving attribution reporting as an
engineering objective. My concern is that a W3C standard should not
unintentionally endorse attribution reporting as a scientifically valid
measure of advertising effectiveness unless the specification is much
clearer about what the API can and cannot establish.

Thanks,

Rick Bruner
CEO, Central Control, Inc.
Newsletter <https://www.centralcontrol.com/newsletter> |
***@***.*** | www.centralcontrol.com
Turn Ad Performance Certainty Into Market Advantage

Free whitepaper: How to design large-scale geographic experiments for
incrementality testing
<https://www.centralcontrol.com/how-to-guide-georct?k=2693226e44>

bmcase:
@rickcentralcontrolcom   you don't see the changes in the spec yet because the PR has not been merged yet.  If you look at this preview link https://pr-preview.s3.amazonaws.com/w3c/attribution/pull/441.html#limitations you can see the new limitations section.

rickcentralcontrolcom:
Thanks, Ben.

I found the preview and have now read the new "Limitations and Successful
Use" section.

I think this is a substantial improvement to the draft. In particular, I
appreciate that the text now explicitly acknowledges that attribution can
create a false impression of advertising efficacy and that randomized
control trials (incrementality experiments) are necessary to measure causal
effects.

That addresses a significant part of my concern.

However, after reading the new section in the context of the rest of the
document, I think there remains a consistency issue.

The new language correctly distinguishes between attribution (which
measures associations) and experimentation (which measures causal effects).
Yet other parts of the document still describe attribution in terms that
appear to imply causal conclusions, such as learning what advertising
"works best," identifying "effective advertising," or helping advertisers
determine which actions lead to desired outcomes.

If the document now acknowledges that attribution alone cannot establish
causality, I think it would be worth reviewing the remaining text for
terminology that may unintentionally suggest otherwise.

More broadly, my concern is less about any specific implementation detail
and more about avoiding a common and consequential misunderstanding: the
assumption that observed associations between ad exposures and outcomes are
evidence that the advertising caused those outcomes. The new section is a
welcome step in that direction, but I believe the same principle should be
reflected consistently throughout the specification.

Thank you for taking the issue seriously and for adding the new material.

Thanks,

Rick Bruner
CEO, Central Control, Inc.
Newsletter <https://www.centralcontrol.com/newsletter> |
***@***.*** | www.centralcontrol.com
Turn Ad Performance Certainty Into Market Advantage

Free whitepaper: How to design large-scale geographic experiments for
incrementality testing
<https://www.centralcontrol.com/how-to-guide-georct?k=2693226e44>

AramZS:
We will merge this on Monday unless there are any reasonable objections.

bmayd:
I want to add a voice to the concerns articulated by @rickcentralcontrolcom and suggest that the document be revised to accurately characterize:

- What browser-based attribution is more broadly.
- What intelligence it can provide and what the limits of that intelligence are.
- What can be reasonably inferred from browser-based attribution more generally.
- Where browser-based attribution fits into larger frameworks that _can_ provide evidence suggestive of advertising effectiveness.
- How the outputs from this API, which are an extremely narrow, purposefully limited and generalized subset of the evidence that can be gathered from browsers, compare with more common browser-derived datasets.
- What the outputs of this API can reliably convey.
- Where the outputs from this API meaningly fit into, and what they contribute to, ad-effectiveness measurement frameworks.

As it stands, the document directly asserts in myriad direct and indirect instances, many outlined by Rick above, that the API enables "the measurement of advertising performance"; that's inaccurate. 

What the API enables is:

- Recording of very dimensionally limited and nondeterministically incomplete data about observable ad-related events.
- Outputs of a subset of those events, correlated using fixed, client-side rulesets constrained by the limits of available data that are unable to account for confounding factors and don't natively incorporate counterfactuals.
- Reporting of limited, noised, aggregate statistics about the correlated outputs.
 
In other words, the output of the API provides an extremely limited, approximate and incomplete snapshot of _what_ happened during a campaign, while by design providing insufficient information to make meaningful inferences about _why_ it happened or adjust for user intent or platform optimization bias.

At best it allows marketers to make limited assumptions about the effectiveness of their targeting and inventory supply in _correlating_ with conversions, but it doesn't provide any insight into if or why the targeting or inventory was or was not effective and it shouldn't imply that it does. The API cannot differentiate between ads that induced a conversion (causation) and those that were merely incidental to it (correlation); it is strictly an observational tracking tool, not an incrementality tool.

Per Rick's comment above, the W3C, via this standard, "should not unintentionally endorse attribution reporting as a scientifically valid measure of advertising effectiveness unless the specification is much clearer about what the API can and cannot establish."

I recognize that an ask to adequately frame this proposal's relationship with rigorous frameworks for ad effectiveness is significant and probably beyond the reasonable scope of this primarily technical standard. If that's the case, I suggest as an alternative that the document be revised to clearly indicate the limits of the API regarding campaign effectiveness measurement, that it include references to resources that can clarify those limits and inform the proper use of this API and that any language suggesting it can directly measure ad effectiveness or campaign performance be modified to clearly indicate it can, at best, provide inputs as part of a properly controlled framework for the assessment of campaign effectiveness.

Suggested introductory abstract:

This document defines a client-side API designed to quantify observable, cross-site associations between digital advertising impressions and subsequent conversion outcomes while preserving user privacy. Crucially, the API does not measure advertising effectiveness or establish a direct causal link between an ad exposure and a user's behavior. Instead, it aggregates data on the chronological sequencing of a limited set of observable ad-related events within a single browser instance. The scope and utility of these insights are inherently bounded by the configurations chosen for measurement—specifically what events are tracked, when lookback windows are applied, and how assignment logic correlates those events. To approximate actual advertising efficacy or incrementality, implementations must supplement these associative baselines with rigorous external experimental designs, such as randomized control trials.

AramZS:
@bmayd Hi! Do you think the current PR's revisions covers those points? I had hoped we made that clear with those revisions and that merging them in would cover these concerns.

rickcentralcontrolcom:
FYI, here's another critique along the same lines of what we've been saying:
https://www.linkedin.com/feed/update/urn:li:activity:7471169938817372160/

I asked ChatGPT to review the doc with an eye for continued causal claims.
Here's what it said:

Yes. The new "Limitations and Successful Use" section is a substantial
improvement. It explicitly says attribution measures associations, not
causality, and that randomized control trials are needed to measure causal
effects.

However, the document still contains numerous passages that would lead a
reasonable reader to conclude that the API helps determine advertising
effectiveness, what works, what performs best, or what causes outcomes.

Here are the strongest examples.
Abstract

"This specifies a browser API for the measurement of advertising
performance."

"The goal is to produce aggregate statistics about how advertising leads to
conversions..."

Both phrases are causal. "Advertising performance" and "how advertising
leads to conversions" imply effectiveness measurement, not mere association.
------------------------------
Background

"the ability to obtain information about the effectiveness of advertising
campaigns"

This explicitly characterizes web attribution as effectiveness measurement.
------------------------------
Background

"Those characteristics made it easier to choose the right audience for
advertising, greatly improving its effectiveness."

Direct claim of improved effectiveness.
------------------------------
Background

"Advertisers seek to place advertising where it will have the most effect
relative to its cost."

Uses "effect" in the causal sense.
------------------------------
Goals

"The measurement of advertising performance creates new cross-site flows of
information."

Again frames the API as measuring performance.
------------------------------
End User Benefit (probably the most problematic section)

"Support for attribution enables more effective advertising..."

That's a very strong claim.
------------------------------

"...helping advertisers understand which ads perform best..."

"Perform best" is exactly the sort of causal inference the limitations
section later says attribution alone cannot establish.
------------------------------

"Connecting that information to outcomes allows an advertiser to learn what
circumstances most often lead to the outcomes they most value."

"Lead to" is causal language.
------------------------------

"When attribution is used effectively, it allows advertisers to spend more
on effective advertising and less on ineffective advertising."

This may be the single strongest overstatement in the document. It says
attribution distinguishes effective from ineffective advertising.
------------------------------

"This lowers the overall cost of advertising relative to the value
obtained."

This asserts a real-world economic benefit that depends on attribution
correctly identifying causal performance.
------------------------------

"Venues for advertising that are better able to show ads that result in the
outcomes that advertisers seek can charge more..."

Again, causation.
------------------------------
Attribution Using Histograms

"grouping by creative ... might be used to learn which creative works best."

The new limitations section says attribution doesn't establish causal
efficacy, yet this sentence still claims it can identify which creative
"works best."
------------------------------
Results and Strategy

"The results that the Attribution API provides can inform decisions about
advertising strategies."

This is milder, but it still implies strategic optimization based on the
output.
------------------------------
Controlled Experiments Example

"Not displaying an advertisement (especially for controlled experiments
that seek to confirm whether an advertising campaign is effective)."

Less problematic because it explicitly refers to experiments, but the
distinction between attribution and experimentation could be clearer.
------------------------------
API Example

"A site that observes a conversion might choose to request the measurement
of the effect of different stored impressions."

This is another direct causal claim. The API measures attributed
associations, not effects.
------------------------------
The Internal Contradiction

The new limitations section now says:

"Some proportion of people who encounter advertising already intend to
follow through... Attributing outcomes to those advertisements creates a
false impression of their efficacy."

and

"To better measure the causal effect of advertising, the use of
(randomized) control trials ... is necessary."

and

"The Attribution API provides aggregate information about the association
between two classes of events: impressions and conversions."

Those statements are largely consistent with your position.

The inconsistency is that the rest of the document still repeatedly tells
readers that attribution:

   -

   measures advertising performance,
   -

   measures effectiveness,
   -

   identifies what ads perform best,
   -

   identifies what creative works best,
   -

   identifies effective vs. ineffective advertising,
   -

   tells advertisers what circumstances lead to outcomes,
   -

   measures the effect of impressions.

In other words, the new limitations section says "this is correlation, not
causation," while many earlier sections continue to describe the API as if
it determines causation. That's the tension that bmayd is pointing to, and
AramZS is essentially asking whether the new limitations section alone is
sufficient to cure that problem.

If the standard were being written as a scientific paper, a reviewer would
probably say the limitations section and the marketing/benefits sections
are not fully aligned.



Thanks,

Rick Bruner
CEO, Central Control, Inc.
Newsletter <https://www.centralcontrol.com/newsletter> |
***@***.*** | www.centralcontrol.com
Turn Ad Performance Certainty Into Market Advantage

Free whitepaper: How to design large-scale geographic experiments for
incrementality testing
<https://www.centralcontrol.com/how-to-guide-georct?k=2693226e44>

bmayd:
@AramZS Thanks for asking directly. I've reviewed the current PR preview. This PR is an improvement and I don't object to it merging — but I don't think merging it should be treated as resolving the underlying concern, and I'd ask for two small additional edits before it does.

The d4ae7d9 commit is welcome but narrow. The passages Rick enumerated remain in the current preview, including the two most direct causal claims:

From the Abstract:

> This specifies a browser API for the measurement of advertising performance. The goal is to produce aggregate statistics about how advertising leads to conversions...

From §1.4 (End-User Benefit):

> Support for attribution enables more effective advertising, largely by helping advertisers understand which ads perform best, and in what circumstances.

The new limitations section directly contradicts both. It states that attribution measures associations, that attributing outcomes to ads can create "a false impression of their efficacy," and that randomized control trials are necessary to measure causal effects. A spec whose abstract claims to measure "advertising performance" and "how advertising leads to conversions," and whose benefits section claims it identifies "which ads perform best," while its limitations section explains it can do none of those things unaided, is internally inconsistent. I'm not asking the group to adopt my view of attribution here — only to make these passages and the rest of the document consistent with what the new section itself says.

Concretely, I'd suggest:

1. Before merge: fix the two passages above. Rick has already drafted suggested replacement language for both in his May 29 comment; either version would do. These are one-line edits.
2. After merge: open a follow-up issue to track the systematic terminology review of the remaining passages Rick enumerated, per his comments and my comment above.

I also want to add context for why this is worth the effort, beyond statistical pedantry. When correlational attribution metrics are taken at face value as measures of effectiveness, they don't just produce noisy answers — they produce systematically *biased* ones, and the bias has a well-understood structural cause.

**The retargeting fallacy and attribution arbitrage**

A system optimized to maximize conversions attributed through correlation learns to identify users with high baseline purchase propensity. Once enough data exists to anticipate a conversion, the economically rational strategy is to follow those users across contexts and serve them ads whose marginal influence is minimal, because the purchase decision has substantially already been made. The more effective a strategy is at identifying consumers already on the path to conversion — which is best accomplished with invasive cross-context tracking — the better it looks under correlational attribution. Attributed CPA and ROAS look excellent; incremental value can be near zero. This is not hypothetical: large-scale randomized experiments have repeatedly found exactly this divergence between attributed and incremental outcomes (e.g., Blake, Nosko & Tadelis's eBay paid-search experiments; Gordon et al.'s comparisons of attribution against experimental lift across Facebook campaigns).

If a W3C standard's framing encourages these metrics to be read as effectiveness measures, the incentive structure it legitimizes is destructive for every constituency:

- **Advertisers** pay a tax on conversions that would have happened anyway, and are pushed into an arms race of handing over first-party customer identity data with the promise it will optimize targeting and the threat that advertiser who don't will lose out to competitors who do.
- **Consumers** — even if the browser mechanism itself is privacy-preserving — face an ecosystem incentivized to collect data *outside* the API to identify users already on a conversion path, then serve them ads primarily to capture attribution credit. They end up overexposed to products they've already decided to buy and underexposed to anything new: the opposite of what advertising is for.
- **Premium publishers** that respect their audiences are penalized. The genuine purchase intent and brand equity that quality sites generate ends up financially credited to low-quality inventory built to harvest attribution credit, and these publishers similarly pressured into surrendering first-party audience data under threat of losing budgets to competitors who do.

**Why this belongs in the W3C's remit**

The W3C's mandate is to guide the web toward its full potential in the interest of users and the health of the ecosystem. By framing an observational, correlation-based mechanism as a tool for measuring "performance" or "effectiveness," the document — however unintentionally — validates a methodology that rewards privacy-invasive tracking, wastes advertiser capital, and starves quality publishers of funding. The limitations section is the right first step; aligning the rest of the document with it is the necessary second one.

Happy to contribute to the follow-up review.

martinthomson:
Maybe @rickcentralcontrolcom, @bmayd, you could help me understand:  If Attribution can be used correctly, to produce aggregated measurements that helps understanding of what advertising is working, why would you object to those claims in the abstract?  The assumption that this is going to be used badly seems implicit in that objection, but the only attribution I've ever seen in practice is exactly the sort of controlled trial.

There might be a well-earned bias from advertising natives that says "attribution is bunkum".  However, for a relative neophyte such as myself, who has only ever dealt with controlled trials, that seems pretty unnecessary.
  
My reaction to Rick's initial pushback was "well, of course if you use this wrong you'll get useless results".

The other pushback I can see is that the claims are overblown.  That any measurement is imperfect and provides only a narrow read on the truth (like the blind men and elephants, if you will).  I don't think that this perspective is disqualifying either, is it?

rickcentralcontrolcom:
You write, "If Attribution can be used correctly, to produce aggregated measurements that helps understanding of what advertising is working..." But that is precisely the point I dispute.

The controlled trial is what tells you what is working. The attribution is incidental. The causal evidence comes from the randomization, not from the attribution mechanism.

I ran advertising research at DoubleClick and Google from 2004-2009, when many of these ideas were first taking shape. Attribution was never originally intended to measure advertising effectiveness. It emerged as a way to allocate fractional credit, and therefore payment, among multiple media companies that touched a conversion path. Over time, the industry began treating it as a measure of ROI, even though it never solved the fundamental causality problem.

If someone sees an ad and later makes a purchase, attribution records that sequence and assigns credit, which was meant to mean payment credit. But it cannot tell us whether the purchase would have happened anyway. That requires a counterfactual, which only a randomized control group can provide.

The W3C proposal improves privacy around attribution reporting, but there is nothing in it that facilitates randomized experiments. In fact, any meaningful experiment would happen entirely outside this mechanism, so claims that it supports experimentation are ill-founded. Whether the randomization is based on user IDs, household IDs, cookies, clean rooms, or geographic regions, the assignment of test and control groups is performed independently of the Attribution API. The proposal may coexist with experiments, but it does not enable them.

That distinction matters because standards have a way of legitimizing ideas. My overarching concern is that the proposal positions attribution reporting in the realm of advertising effectiveness when attribution has never been capable of measuring causal impact.

The danger is not technical. It is conceptual. By having an august body like the W3C be perceived to legitimize and institutionalize the claim that attribution is a tool for understanding what advertising works, the industry risks perpetuating a fallacy that has already distorted media investment for decades. The result is predictable: advertisers are steered toward what is easiest to observe rather than what actually drives incremental business outcomes. That perpetuates waste and misallocation of advertising budgets while slowing the industry's adoption of genuinely causal measurement methods. At the same time, spending becomes increasingly concentrated among the handful of platforms best positioned to generate attribution signals (i.e., the biggest tech giants), to the detriment of thousands of other media companies that are already struggling to survive.

That's why I object to claims that attribution helps determine advertising effectiveness. Attribution measures sequences of events. Experiments measure causal impact.

I laid out my broader concerns about the proposal here:

https://www.adexchanger.com/data-driven-thinking/the-w3c-is-making-a-critical-mistake-about-measuring-advertising-effectiveness/

bmayd:
@martinthomson as I was responding @rickcentralcontrolcom expressed the core concern better than what I was writing — causal evidence comes from the randomization, which determines whether what's measured can be read causally, not the attribution mechanism that does the measuring, and the randomization and assignment of test and control groups is external to this API. The "used correctly" framing is the crux: when measuring causality, the correct use of the API is as a means of counting correlated events within experimental subgroups, at least one of which is a properly defined, randomly selected control group.

Your experience — attribution used inside a controlled trial — is the right approach; the problem is it's the exception. The reality in adtech is the vast majority of digital ad spend is focused exclusively on campaigns consisting only of treatment groups: groups targeting inventory optimized automatically by algorithms trained on simple, observational last-interaction or multi-touch correlation. There are rarely control groups (outside of platforms running discrete lift studies) and almost never on the long tail of inventory. In other words, in the vast majority of cases campaign performance is measured exclusively based on correlation observations derived from a deliberately optimized, highly biased sample, the raw output of which is unfit for use in measuring causality. When the W3C Attribution draft includes statements that directly assert this API is for "the measurement of advertising performance" and produces statistics about "how advertising leads to conversions," and other similar statements, it validates false market assumptions that the API's raw output equals causal impact. Adtech, which has spent the past two decades optimizing to intercept the path to conversion and reporting it as proof of campaign success, will take that endorsement at face value and as justification to perpetuate the practice.

Regarding the blind-men-and-the-elephant framing: if the objection were "all measurement is imperfect and gives a narrow read," I'd agree it wouldn't be disqualifying and I wouldn't be raising it. That's not the issue; imperfection is about *fidelity* — a noisy or partial read on the right quantity. This is about *identity* — association and causation are categorically different and no matter how precise and complete observational attribution data is, it doesn't measure causal effect. A caveat mentioning this in any one section isn’t sufficient when the rest of the document strongly implies, or outright states, the API measures performance.

And to your suggestion "the objection assumes it'll be used badly", it doesn't assume bad faith. The *default* use of an attribution API is observational — impressions matched to conversions, no control arm — and as the limitations section states, that use creates "a false impression of [advertising’s] efficacy." The concern is the ordinary case the spec itself flags, not a hypothetical abuse. Nor is the implication "attribution is bunkum": attribution is genuinely useful for what it measures — which placements, creatives, and contexts are *associated* with conversions, i.e. diagnostics, optimization inputs, and counting inside an experiment. The objection is to *labeling* that association "effectiveness" or "performance" and implying the API measures more than it can. 

As suggested in previous posts, the instances where language clearly asserts or implies that the API measures performance ought to be revised, if not before this PR is merged, then in a follow-up terminology review – happy to help with that.

martinthomson:
This is helpful.  My initial reaction to some of the feedback from Rick was that this was an overreaction.  That is, the claims in the document were about the value of measurement more generally, which I think is still defensible.  But your comments have highlighted that this was being directly linked to attribution, rather than saying that attribution provides information as part of a larger measurement strategy that can then contribute to those outcomes.  I've made a few more changes, notably in the introduction, goals, and user benefit sections.  I hope those are in the right direction.

I still think it is necessary to take some liberties with the user benefit section, but I hope that the line is clearer now: attribution -> measurement -> better advertising.

api.bs
```diff
@@ -155,19 +173,20 @@ path:images/value.svg
 Participation in an [=attribution=] measurement system
 would comprise a secondary cost to Web users.
 
-Support for attribution enables more effective advertising,
-largely by helping advertisers understand which ads perform best,
-and in what circumstances.
-Those circumstances might include
-the time and place that the ad is shown,
+Support for attribution can enables more effective advertising,
```
bmayd: Should be `enable` instead of `enables`.

```suggestion
Support for attribution can enable more effective advertising,
```

bmayd:
@martinthomson Good updates which improve things overall. I found a typo ("enables" should be "enable") I flagged in a comment. I still think a terminology review is in order before finalizing, some of the things Rick pointed out are still outstanding, but don't think it should hold this up so @AramZS I'm OK with merging this.

As the limitations section now observes, "the web platform presently does not have a facility for providing consistent allocation of users into control and treatment groups across different sites." I've been working on a proposal for filling that gap so the API can support incrementality measurement. I'll post in separate issue a suggestion for a privacy-preserving way to provide that facility, and I'll link it here when it's up — intended as the constructive complement to this PR, not anything that should slow it down.

bmayd:
I posted issue #451 with suggested additions for incrementality measurement support.

AramZS:
WG has consensus to merge this PR in. Let's move forward and iterate where needed.

PR #458, opened 2026-07-08
Simplify cross-budget-deduction state tracking

apasel422:
The "deduct privacy and safety budgets" algorithm is invoked exactly once per epoch in the chosen range, so it is unconditionally true that each epoch's global budget is deducted from at most once.

Similarly, a given impression belongs to a single epoch, so there is no need to maintain a set keyed by epoch across invocations of that algorithm: We only need to remember whether a given impression *site* has already had its quota modified in the current epoch.

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/ppa-api/pull/458.html" title="Last updated on Jul 13, 2026, 5:11 PM UTC (50c1dbd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/458/a761634...apasel422:50c1dbd.html" title="Last updated on Jul 13, 2026, 5:11 PM UTC (50c1dbd)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -1205,8 +1205,7 @@ given a [=privacy budget key=] |key|,
 integer |value|,
 integer |maxValue|,
 boolean |isSingleEpoch|,
-integer |l1Norm|,
-and a [=set=] of [=privacy budget keys=] |deductedImpressionQuotas|:
+and an integer |l1Norm|:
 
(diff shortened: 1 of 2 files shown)


csharrison (approved):

bmcase:
yes, I think this works. There may be multiple impressions in the same epoch with the same impression site, but you will catch that by just looking if the site has been deducted from in that epoch in the range.

bmcase (approved):

PR #461, opened 2026-07-08
Cleanly split single-epoch and multi-epoch attribution handling

### LINKED ISSUE #397: [clean up] improve how we pass |isSingleEpoch| around
from [comment](https://github.com/w3c/attribution/pull/309#discussion_r3013363219): Can we save the |isSingleEpoch| variable and have a method on an impression set that answers yes or no to [=is single epoch=]?

apasel422:
This makes it clear that the L1 norm computation is irrelevant for the latter, and inherently avoids the redundant common matching logic and histogram creation for the former.

Fixes #397.

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/ppa-api/pull/461.html" title="Last updated on Jul 30, 2026, 3:21 PM UTC (2926c2f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/461/3362d5c...apasel422:2926c2f.html" title="Last updated on Jul 30, 2026, 3:21 PM UTC (2926c2f)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -1204,16 +1204,11 @@ given a [=privacy budget key=] |key|,
 [[WEBIDL#idl-double|double]] |epsilon|,
 integer |value|,
 integer |maxValue|,
-boolean |isSingleEpoch|,
-and an integer |l1Norm|:
-
-1.  Let |l1NormSensitivity| be |l1Norm| if |isSingleEpoch|, 2 * |value| otherwise.
+and an integer or null |singleEpochL1Norm|:
 
 1.  Let |valueSensitivity| be 2 * |value|.
 
 1.  Let |noiseScale| be 2 * |maxValue| / |epsilon|.
-
-1.  Let |l1NormDeductionFp| be |l1NormSensitivity| / |noiseScale|.
 
 1.  Let |valueDeductionFp| be |valueSensitivity| / |noiseScale|.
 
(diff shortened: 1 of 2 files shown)


api.bs
```diff
@@ -1225,16 +1220,22 @@ and an integer |l1Norm|:
     proportional to |maxValue| / |epsilon|
     is added to the aggregated histogram.
```
martinthomson: It might be worth revising this note and including it only for the conditional arm.  That is, the main line uses replacement for its sensitivity analysis because of the potential for cascading effects between epochs, whereas the single epoch has no need for that.

api.bs
```diff
@@ -2059,20 +2058,31 @@ To <dfn>do attribution and fill a histogram</dfn>, given
...
     1.  [=Assert=]: |l1Norm| is less than or equal to |options|' [=validated conversion options/value=].
 
+    1.  Let |key| be a [=privacy budget key=] whose items are |currentEpoch| and |topLevelSite|.
+
+    1.  Let |budgetAndSafetyOk| be the result of invoking [=deduct privacy and safety budgets=]
+        with |key|, |impressions|,
+        |options|' [=validated conversion options/epsilon=],
+        |options|' [=validated conversion options/value=],
+        |options|'s [=validated conversion options/max value=],
+        and |l1Norm|.
```
martinthomson: With so many arguments, it's a bit hard to track what these are, at least when they are not already named to match the argument name.  Especially for the "null" below.  Is there a better convention here?

martinthomson (approved):

api.bs
apasel422: Done.

PR #466, opened 2026-07-15
Remove erroneous assumption that a key exists in global privacy budget store

apasel422:
In the specification, the global privacy budget store was only written to during the deduction, but the first time the epoch's global budget is ever modified, the key doesn't exist, so it must be initialized.

In the simulator, Map.get returns undefined, which is coerced into NaN as a result of the subtraction, causing NaN to be written into the map during the subsequent Map.set call. Later global budget checks for the same epoch would succeed because `valueDeduction > currentGlobalValue` returns false when currentGlobalValue is NaN. Rewriting that check as `!(valueDeduction <= currentGlobalValue)` correctly causes the tests to fail without the fix in this commit.

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/ppa-api/pull/466.html" title="Last updated on Jul 15, 2026, 2:26 PM UTC (f8965f7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/466/3c1615c...apasel422:f8965f7.html" title="Last updated on Jul 15, 2026, 2:26 PM UTC (f8965f7)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -1255,7 +1255,12 @@ and an integer |l1Norm|:
 
     1.  Let |epoch| be the [=epoch index=] component of |key|.
 
-    1.  Decrement [=global privacy budget store=]\[|epoch|] by |valueDeduction|.
+    1.  Let |currentGlobalValue| be the result of [=map/get|getting the value=]
+        of |epoch| in the [=global privacy budget store=]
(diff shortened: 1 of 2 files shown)


bmcase (approved):
Looks good to me

PR #467, opened 2026-07-15
Surface errors when clearing an empty site set

### LINKED ISSUE #293: Simulator throws an exception when forgetting state with the empty set
Due to https://github.com/w3c/attribution/blob/74fa8b63ff8e640b8a0add11411aa46b1d1fe535/impl/src/backend.ts#L625-L627

The exception itself makes sense in terms of business logic, but the issue isn't propagated to the UI, so the user has no indication of why their intended operation had no effect.

ckdwns9121:
Fixes #293.

## What changed

- replace the generic assertion for an empty site set with a descriptive `RangeError`
- catch clear-state failures in the simulator and show the result in an accessible live output
- add regression coverage for the error type and message

## Why

Clearing site state with no sites and without selecting **Forget Visits** is invalid, but the exception previously escaped the submit handler. The operation appeared to do nothing and gave the user no explanation.

An empty site set remains valid when **Forget Visits** is selected because that operation intentionally clears all visits.

## Validation

- `npm run build`
- `npm test` — 126 tests passed
- `npm run lint`
- `npm run pretty:check`
- `npm run pack`
- headless Chrome submission of the affected form, confirming the descriptive error appears in the live output

AI assistance disclosure: Codex assisted with investigation, implementation, and testing. I reviewed and verified the changes.

### DIFF
--- a/impl/index.html
+++ b/impl/index.html
@@ -139,6 +139,7 @@
             >
             <button>Clear Site Data as User</button>
           </line>
+          <output aria-live="polite"></output>
         </form>
         <form id="clear-as-site" class="simple">
           <line>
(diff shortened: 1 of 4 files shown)


ckdwns9121:
Closing this draft before review because I’m stepping back from #293. Sorry for the noise.

PR #478, opened 2026-08-11
Deal with localhost on implicit values

### LINKED ISSUE #456: Inconsistent treatment of localhost
#421 blocked the use of `localhost` within the [`parse a site`](https://w3c.github.io/attribution/#parse-a-site) algorithm, which is used to validate the impression/conversion sites/callers fields of options passed to the `save impression` and `measure conversion` algorithms.

It did *not* prevent the use of `localhost` as the top-level site or intermediary, as the [`obtain the implicit API inputs`](https://w3c.github.io/attribution/#obtain-the-implicit-api-inputs) algorithm was not changed, despite the following non-normative note below `parse a site`:

https://github.com/w3c/attribution/blob/c18f38afec899b402cb35d20db8ac862ff1ba811/api.bs#L1093-L1099

Subsequently #428 modified the simulator to prevent all uses of `localhost`, counter to the current state of the specification.

IIUC, the intent of #421 was to block all uses, as it doesn't really make sense to allow `localhost` as the top-level site or intermediary if the options cannot specifically permit them, so we should update the `obtain the implicit API inputs` algorithm to be consistent with the simulator and enforce this.

martinthomson:
(This sat on a branch for a whole month.  Mostly because I'm not sure if this is the right sort of check, because it expands in a very awkward way.)

I've expanded the scope of this a little to include literal IP addresses as well, after I realized that we weren't able to catch a bunch of problematic names through their use of IP addresses.  This is because of awkward things like `[::1]`, but also the wide array of ways of spelling loopback as an IPv4 address.  There are also a bunch of non-unique IP addresses that could end up being used to poison things.

The check in code is pretty small, but it's not necessarily obvious.

The knock-on effect was that it didn't make sense to use SyntaxError for the case where implicit inputs are on loopback.  I switched all of the error markings to NotAllowed instead.

Closes #456.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/pull/478.html" title="Last updated on Aug 13, 2026, 12:08 AM UTC (716acb1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/478/b243f24...716acb1.html" title="Last updated on Aug 13, 2026, 12:08 AM UTC (716acb1)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -1738,7 +1738,8 @@ and a [=moment=] |now|:
 <div algorithm>
 The <dfn method for=Attribution>saveImpression(|options|)</dfn> method steps are:
 
-1.  Let |implicitInputs| be the result of [=obtaining the implicit API inputs=] from [=this=].
+1.  Let |implicitInputs| be the result of [=obtaining the implicit API inputs=] from [=this=],
+    returning [=a promise rejected with=] any thrown reason.
 1.  [=Assert=]: |implicitInputs| is not null.
 1.  Return the result of running [=save an impression=] with |options| and |implicitInputs|.
 
@@ -1889,7 +1890,8 @@ asynchronously, queuing work on the <dfn>Attribution [=task source=]</dfn>.
 <div algorithm>
 The <dfn method for=Attribution>measureConversion(|options|)</dfn> method steps are:
 
-1.  Let |implicitInputs| be the result of [=obtaining the implicit API inputs=] from [=this=].
+1.  Let |implicitInputs| be the result of [=obtaining the implicit API inputs=] from [=this=],
+    returning [=a promise rejected with=] any thrown reason.
 1.  [=Assert=]: |implicitInputs| is not null.
 1.  Return the result of running [=measure a conversion=] with |options| and |implicitInputs|.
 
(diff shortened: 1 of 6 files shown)


api.bs
```diff
@@ -1981,6 +1860,27 @@ optional [=origin=] (default null):
...
+returning a [=site=]:
+
+1.  If |origin| is not a [=tuple origin=],
+    throw a  {{"NotAllowedError"}} {{DOMException}}.
+1.  Let |host| be the  [=host=] component of |origin|.
+1.  If |host| is "`localhost`",
+    if |host| [=code unit suffix|ends with=] "`.localhost`",
+    if |host| [=code unit prefix|starts with=] U+005B ([),
+    or if |host| [=ends in a number=],
+    throw a  {{"NotAllowedError"}} {{DOMException}}.
```
```suggestion
    throw a {{"NotAllowedError"}} {{DOMException}}.
```

api.bs
apasel422: Just clarifying intent here: As written, if `measureConversion` is invoked on `localhost`, `obtaining the implicit API inputs` will throw an exception, which is presumably propagated at this point of the algorithm, meaning that in JS it will be thrown directly, rather than produced as part of the promise resolution that the `measure a conversion algorithm` does below, making it different from all the other failure modes.

We probably want to return a rejected promise instead for consistency, right?

The same goes for the `saveImpression` algorithm, and we probably want to explicitly handle the failure in Steps 5-6 of the `handle Attribution headers` algorithm to make it clear for implementers how they're supposed to proceed.

apasel422 (changes requested):
Can we add a couple of new test cases to https://github.com/w3c/attribution/blob/main/impl/e2e-tests/save-impression-localhost.json and https://github.com/w3c/attribution/blob/main/impl/e2e-tests/measure-conversion-localhost.json reflecting the new behavior? Ideally these would cover the implicit inputs in general, and specifically the IP address and `.localhost` suffix stuff.

api.bs
martinthomson: Yep, that's a good point.  This should return a rejected promise for the main functions and ignore the attempt to invoke the API when dealing with headers (logging notwithstanding).  Modern code with `await` won't care, but it's best to be neat.

apasel422 (approved):

PR #484, opened 2026-08-20
It's not central DP

### LINKED ISSUE #479: Which DP model?
I hope that @npdoty doesn't mind me quoting his email here, but he identifies an editorial error.

---

> This is due to the effect of aggregation on the ability of sites to extract information about individuals from aggregates. This is especially true for central differential privacy, which is the mathematical basis for the privacy design used in this specification.

This is confusing text to readers like me, as the spec also says:

> This attribution design uses a form of differential privacy called individual differential privacy. In this model, user agents are each separately responsible for ensuring that they limit the information that is contributed.

Is the basis of the design central differential privacy or individual
differential privacy?

---

It's individual, though the confusing statement is still directionally correct, in that the noise is centrally applied.  Easy to correct, I think.

martinthomson:
The noise application part is basically central, but this is misleading. The individual model we use doesn't really care about where noise is added, so we still need to be clear we are adding noise at the aggregation service.

This was hard text to touch, as it is in a tricky part of the spec.  I'd appreciate careful review.

Closes #479.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/pull/484.html" title="Last updated on Aug 28, 2026, 10:03 AM UTC (ea83220)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/attribution/484/91ef174...ea83220.html" title="Last updated on Aug 28, 2026, 10:03 AM UTC (ea83220)">Diff</a>

### DIFF
--- a/api.bs
+++ b/api.bs
@@ -227,13 +227,27 @@ when the group that participates is larger.
 This is due to the effect of aggregation on
 the ability of sites to
 extract information about individuals from aggregates.
-This is especially true for central [[#dp|differential privacy]],
-which is the mathematical basis for the privacy design used
-in this specification.
 
-Larger cohorts of participants also produce more representative--
-and therefore more useful--
-statistics about the advertising that is being measured.
+This is especially true for the [[#dp|differential privacy]]
+model used in this specification.
+Noise is added during aggregation as a single quantity
+that is proportional to the amount that an individual contributes.
+Increased participation reduces the degree to which
+this noise affects the usefulness of the results.
+This allows the noise to be large
+relative to the values that any single person might contribute.
+
+Alternative models that rely on noise added to individual contributions
+produce more noise for the same privacy goal
+as the magnitude of noise increases with more contributions.
+The result is a worse balance of privacy and utility.
+To achieve the same utility in these systems,
+either looser privacy protections
+or measurement of significantly larger populations
+is needed.
+
+Larger cohorts of participants also produce more representative statistics,
+which are inherently more useful.
 
 If attribution is justified,
 both these factors motivate the enablement of attribution for all users.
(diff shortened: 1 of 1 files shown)


api.bs
```diff
@@ -227,13 +227,27 @@ when the group that participates is larger.
...
-which is the mathematical basis for the privacy design used
-in this specification.
 
-Larger cohorts of participants also produce more representative--
-and therefore more useful--
-statistics about the advertising that is being measured.
+This is especially true for the [[#dp|differential privacy]]
+model used in this specification.
+Noise is added during aggregation as a single quantity
+that is proportional to the amount that an individual contributes.
```
bmcase, line 234: This is a bit subtle in that the noise added is always sufficient for what was requested to be spent as budget from your device, but one user may have a larger contribution toward the aggregate than other users in the aggregate if that user had more request from their device.  

For example: 
suppose there are two separate conversion requests to a user's device each asking to spend eps/2 budget for a Delta sensitivity query.   The device deducts privacy budget for each as (eps/2)/(individual sensitivity/ Delta).   At max individual sensitivity/ Delta = 1 and all of the devices eps budget is spent.  

Then both of these get aggregated in the same batch with Lap(Delta/ (esp/2) ) noise applied.  This user actually contributes 2*Delta toward the aggregate, but that is okay because they spent all their eps budget where as other users whose contribution to the aggregate is Delta spent only half of their budget.  

so I would suggest something like the following. Plus separately adding a note about how the size of the aggregate doesn't impact the theoretical DP protection provided.   
```suggestion
Noise added during aggregation is sufficient to protect the privacy of any individual contributing to the aggregate in accordance with how much privacy budget was spent from that individual's device.  The DP guarantees do not change based on the aggregation group size; however, aggregating with larger groups can provide its own additional privacy benefit beyond what can be measured by DP. 
```

bmcase (commented):
comment on describing how much noise is added

bmcase (approved):

api.bs
martinthomson: Does this work?

```suggestion
Noise is added during aggregation as a single quantity
that is based on the amount of privacy budget
that has been consumed from each individual
that contributes to the aggregate.
Differential privacy guarantees do not change
based on the number of individuals who participate.
```

bmcase: @martinthomson  yes that looks good

